PR #2, opened 2017-11-09
HMAC authenticaion method

OndraM:
Only a basic building block of client internals.

Will be used as [custom authentication method](http://docs.php-http.org/en/latest/message/authentication.html#implement-your-own) of HTTPlug [Authentication plugin](http://docs.php-http.org/en/latest/plugins/authentication.html).

Something like
```php
new PluginClient(
    HttpClientDiscovery::find(),
    [
        new AuthenticationPlugin(new HmacAuthentication($this->apiKey)),
    ]
);
```
(This will be part of Matej\Client internals, so end-clients will not have to set it up.)

### DIFF
--- a/src/Http/HmacAuthentication.php
+++ b/src/Http/HmacAuthentication.php
@@ -1,0 +1,61 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Http;
+
+use Http\Message\Authentication;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Use api key to sign request with hmac_timestamp and hmac_sign query parameters.
+ */
+class HmacAuthentication implements Authentication
+{
+    /** @var string */
+    private $apiKey;
+
+    public function __construct(string $apiKey)
+    {
+        $this->apiKey = $apiKey;
+    }
+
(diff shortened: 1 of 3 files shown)


src/Http/HmacAuthentication.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public function authenticate(RequestInterface $request): RequestInterface
+    {
+        $params = $this->addAuthenticationParams($request);
+
+        return $this->createRequestWithParams($request, $params);
+    }
+
+    private function addAuthenticationParams(RequestInterface $request): array
```
jirinovak: Maybe `getAuthenticationParams` or `createAuthenticationParams` would be better names.

foglcz (approved):

MartinKristof (approved):

src/Http/HmacAuthentication.php
OndraM: I've been thinking about this one before and also considered both of those suggestions. But I didn't like them as well - the params are actually being added to params already present in Request. So they are not being created, nor just "get" (get = no side effects). But I don't like `addAuthenticationParams` either.

So what about `addAuthenticationToParamsFromRequest($request)`?

PR #4, opened 2017-11-10
Exception handling plugin

OndraM:
Also part of the internal HTTP layer. Will transfer error responses from Matej API to custom Exceptions.

It will be used internally in the library as another HTTPlug request plugin (http://docs.php-http.org/en/latest/plugins/build-your-own.html).

```php
$pluginClient = new PluginClient(
    $this->getHttpClient(),
    [
        // ... other plugins, like AuthenticationPlugin
        new ExceptionPlugin(),
    ]
);
```

### DIFF
--- a/src/Exception/AuthorizationException.php
+++ b/src/Exception/AuthorizationException.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Exception;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Exception thrown when request authorization fails.
+ */
+class AuthorizationException extends RequestException
+{
+    public static function createFromRequestAndResponse(
+        RequestInterface $request,
+        ResponseInterface $response,
+        \Throwable $previous = null
+    ): self {
+        $responseData = json_decode($response->getBody()->getContents());
+
+        $message = sprintf(
+            'Matej API authorization error for url "%s"%s',
+            $request->getRequestTarget(),
+            isset($responseData->message) ? ' (' . $responseData->message . ')' : ''
+        );
+
+        return new static($message, $request, $response, $previous);
+    }
+}
--- a/src/Http/Plugin/ExceptionPlugin.php
+++ b/src/Http/Plugin/ExceptionPlugin.php
@@ -1,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Http\Plugin;
+
+use Fig\Http\Message\StatusCodeInterface;
+use Http\Client\Common\Plugin;
+use Http\Promise\Promise;
+use Lmc\Matej\Exception\AuthorizationException;
+use Lmc\Matej\Exception\RequestException;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+final class ExceptionPlugin implements Plugin
+{
+    public function handleRequest(RequestInterface $request, callable $next, callable $first): Promise
+    {
+        /** @var Promise $promise */
+        $promise = $next($request);
+
+        return $promise->then(function (ResponseInterface $response) use ($request) {
+            return $this->transformResponseToException($request, $response);
+        });
+    }
+
+    private function transformResponseToException(RequestInterface $request, ResponseInterface $response)
+    {
+        $responseCode = $response->getStatusCode();
+
+        if ($responseCode === StatusCodeInterface::STATUS_UNAUTHORIZED) {
+            throw AuthorizationException::createFromRequestAndResponse($request, $response);
+        }
+
+        if ($responseCode >= 400 && $responseCode < 600) {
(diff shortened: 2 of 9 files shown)


src/Exception/AuthorizationException.php
```diff
@@ -0,0 +1,28 @@
...
+ * Exception thrown when
+ */
+class AuthorizationException extends RequestException
+{
+    public static function createFromRequestAndResponse(
+        RequestInterface $request,
+        ResponseInterface $response,
+        \Throwable $previous = null
+    ) {
+        $responseData = json_decode($response->getBody()->getContents(), false);
```
jirinovak: false is the default

src/Exception/AuthorizationException.php
```diff
@@ -0,0 +1,28 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Exception;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Exception thrown when
```
jirinovak: Ok, when? :-)

src/Http/Plugin/ExceptionPlugin.php
```diff
@@ -0,0 +1,40 @@
...
+    private function transformResponseToException(RequestInterface $request, ResponseInterface $response)
+    {
+        $responseCode = $response->getStatusCode();
+
+        if ($responseCode === StatusCodeInterface::STATUS_UNAUTHORIZED) {
+            throw AuthorizationException::createFromRequestAndResponse($request, $response);
+        }
+
+        if ($responseCode >= 400 && $responseCode < 600) {
+            // TODO: use more specific exceptions
```
jirinovak, line 34: Maybe we could create issue and link it here.

tests/Http/Plugin/ExceptionPluginTest.php
```diff
@@ -0,0 +1,77 @@
...
+class ExceptionPluginTest extends TestCase
+{
+    /** @var callable */
+    protected $emptyFunction;
+
+    /** @before */
+    public function init(): void
+    {
+        $this->emptyFunction = function (): void {
+        };
```
jirinovak: same line as `{`?

tests/Http/Plugin/ExceptionPluginTest.php
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * @return array[]
+     */
+    public function provideErrorStatusCodes(): array
+    {
+        return [
+            'HTTP 400' => [StatusCodeInterface::STATUS_BAD_REQUEST, RequestException::class],
+            'HTTP 401' => [StatusCodeInterface::STATUS_UNAUTHORIZED, AuthorizationException::class],
+            'HTTP 404' => [StatusCodeInterface::STATUS_NOT_FOUND, RequestException::class],
+            'HTTP 500' => [StatusCodeInterface::STATUS_INTERNAL_SERVER_ERROR, RequestException::class],
```
jirinovak, line 78: What about testing `< 600` boundary.

src/Exception/AuthorizationException.php
OndraM: That's a question!

Fixed :)

src/Http/Plugin/ExceptionPlugin.php
OndraM: There is/will be more of this, the whole repo is TODO right now :-D. I will make sure fix all TODOs  before the 1.0 version (or do issues where needed), though.

tests/Http/Plugin/ExceptionPluginTest.php
OndraM: Its (unfortunately, in this case) forbidden by PSR-2 and causes php-cs-fixer to cry.

> Opening braces for control structures MUST go on the same line, and closing braces MUST go on the next line after the body.

¯\\\_(ツ)_/¯

tests/Http/Plugin/ExceptionPluginTest.php
OndraM: There is no HTTP 600 :). Byt I added the check for (imaginary) HTTP 599 code as well.

jirinovak (approved):

tests/Http/Plugin/ExceptionPluginTest.php
foglcz: Technically, function isn't a [control structure](http://php.net/manual/en/language.control-structures.php) (edit: I think :smile:)

In php-cs-fixer, they mention `allow_single_line_closure` (defaults to false). Maybe that'd help?

foglcz (approved):

tests/Http/Plugin/ExceptionPluginTest.php
OndraM: I would sey they are, but this is kinda gray zone. However, nice finding about the allow_single_line_closure setting! I'm for adding this to configuration and us the one-liner. This newline makes it surely less readable for no reason.

PR #7, opened 2017-11-13
Response decoding

OndraM:
Response JSON is decoded to `Response` object, which then contains array of `CommandResponses` - responses to individual commands which were send as part of request batch.

The `Response` could later maybe implement \Iterator, \ArrayAccess or maybe even \SplObjectStorage for simpler manipulation with the Command Responses. (see https://github.com/lmc-eu/matej-client-php/issues/8)

Also note the compatibility layer for plain string command responses. It will be removed right after https://jira.int.lmc.cz/browse/RAD-643 is released.

### DIFF
--- a/src/Model/CommandResponse.php
+++ b/src/Model/CommandResponse.php
@@ -1,0 +1,55 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model;
+
+use Lmc\Matej\Exception\InvalidDomainModelArgumentException;
+
+/**
+ * Response to one single command which was part of request batch.
+ */
+class CommandResponse
+{
+    const STATUS_OK = 'OK';
+    const STATUS_ERROR = 'ERROR';
+    const STATUS_SKIPPED = 'SKIPPED';
+
+    /** @var string */
+    private $status;
+    /** @var string */
+    private $message;
+    /** @var array */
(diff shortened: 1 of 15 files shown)


foglcz (approved):

src/Model/CommandResponse.php
```diff
@@ -0,0 +1,56 @@
...
+    const STATUS_SKIPPED = 'SKIPPED';
+
+    /** @var string */
+    private $status;
+    /** @var string */
+    private $message;
+    /** @var array */
+    private $data = [];
+
+    private function __construct()
```
mvondr, line 23: Empty constructor?

davidkmenta: A private empty constructor means that this class cannot be instantiated.

src/Model/CommandResponse.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+    }
+
+    public static function createFromRawCommandResponseObject(\stdClass $rawCommandResponseObject): self
+    {
+        $commandResponse = new self();
```
davidkmenta: This should be after the condition below

davidkmenta (dismissed):

mvondr (approved):

PR #10, opened 2017-11-14
Request manager to handle HTTP layer

OndraM:
`RequestManager` takes prepared instance of `Lmc\Matej\Model\Request`, executes it (adding HmacAuth etc. to the HttpClient) it and processes the Http response to internal domain model (`Lmc\Matej\Model\Response`).

Usage example:
```php
$matejRequest = new \Lmc\Matej\Model\Request('/item-properties', 'PUT', [[...], [...]]);

$matejResponse = $requestManager->sendRequest($matejRequest);

// => instance of Lmc\Matej\Model\Response
```

This is again part of API client internal HTTP layer. The example above *will not* be used directly in application code - the Request instance will ve prepared using builder objects (which will be done soon).

### DIFF
--- a/src/Http/RequestManager.php
+++ b/src/Http/RequestManager.php
@@ -1,0 +1,126 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Http;
+
+use Http\Client\Common\Plugin\AuthenticationPlugin;
+use Http\Client\Common\PluginClient;
+use Http\Client\HttpClient;
+use Http\Discovery\HttpClientDiscovery;
+use Http\Discovery\MessageFactoryDiscovery;
+use Http\Message\MessageFactory;
+use Lmc\Matej\Http\Plugin\ExceptionPlugin;
+use Lmc\Matej\Model\Request;
+use Lmc\Matej\Model\Response;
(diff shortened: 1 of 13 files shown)


foglcz (approved):

davidkmenta (approved):

src/Http/RequestManager.php
```diff
@@ -0,0 +1,126 @@
...
+use Http\Discovery\MessageFactoryDiscovery;
+use Http\Message\MessageFactory;
+use Lmc\Matej\Http\Plugin\ExceptionPlugin;
+use Lmc\Matej\Model\Request;
+use Lmc\Matej\Model\Response;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Encapsulates HTTP layer, ie. request/response handling.
+ * This class should typically not be used directly - its supposed to be called internally from `Matej` class.
```
kubasimon: "This class should not be typically used directly." - better word order

kubasimon (approved):

src/Http/RequestManager.php
OndraM: Adjusted, thanks

PR #11, opened 2017-11-15
Commands to setup and update database items

OndraM:
Matej endpoints accepts set of specific Commands.

They will be used when building request to Matej. Unlike previous PRs, **these classes will actually be used by the end application**, so make sure their API fits your needs.

For example the final usage from application point of view will be like:
```php
// to setup new item properties in the database
$response = $matej->request()
    ->itemPropertiesSetup()
    ->addProperty(ItemPropertySetup::timestamp('valid_from'))
    ->send();
```

Or
```php
// to send events
$response = $matej->request()
    ->events()
    ->addProperty(ItemProperty::create('.... adId', ['valid_to' => 1510756958])) // update properties of item (ad)
    ->addInteraction('adId', 'userId', 'interactionType', ...) // to submit interactions - not part of this PR
    ->send();
```

The `request()` method, `itemPropertiesSetup()` and `send()` etc. will be part of some future PR - this PR is just about first two Command types.

### DIFF
--- a/src/Model/Command/ItemProperty.php
+++ b/src/Model/Command/ItemProperty.php
@@ -1,0 +1,41 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+/**
+ * Command to save different item content properties to Matej.
+ */
+class ItemProperty extends AbstractCommand
+{
+    /** @var string */
+    private $itemId;
+    /** @var array */
+    private $properties;
+
+    private function __construct(string $itemId, array $properties)
+    {
+        // TODO: assert itemId format
+
+        $this->itemId = $itemId;
+        $this->properties = $properties;
+    }
+
+    public static function create(string $itemId, array $properties = []): self
+    {
+        return new static($itemId, $properties);
+    }
+
+    protected function getCommandType(): string
+    {
+        return 'item-properties';
+    }
+
+    protected function getCommandParameters(): array
+    {
+        $parameters = $this->properties;
+
+        $parameters['item_id'] = $this->itemId;
+
+        return $parameters;
+    }
+}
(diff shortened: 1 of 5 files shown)


tests/Model/Command/ItemPropertyTest.php
```diff
@@ -0,0 +1,45 @@
...
+    public function provideProperties(): array
+    {
+        return [
+            'No item properties' => [[], ['item_id' => 'exampleItemId']],
+            'One item property' => [['date' => 1510756952], ['date' => 1510756952, 'item_id' => 'exampleItemId']],
+            'Multiple item properties' => [
+                ['item1' => 'value1', 'item2' => 'value2'],
+                ['item1' => 'value1', 'item2' => 'value2', 'item_id' => 'exampleItemId'],
+            ],
+            'Shod not allow to override item_id' => [
```
kubasimon: typo: should

kubasimon (approved):

tests/Model/Command/ItemPropertyTest.php
OndraM: Fixed

MarketaSebkova (approved):

davidkmenta (approved):

src/Model/Command/ItemProperty.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        // TODO: assert itemId format
+
+        $this->itemId = $itemId;
+        $this->properties = $properties;
+    }
+
+    public static function create(string $itemId, array $properties = []): self
+    {
+        return new self($itemId, $properties);
```
foglcz: Maybe `return new static();` ?

foglcz (approved):

src/Model/Command/ItemProperty.php
OndraM: True, would be better. Thanks

PR #13, opened 2017-11-16
Adapt to latest API changes

OndraM:
Adapt to changes from Matej release on Wednesday evening .

https://jira.int.lmc.cz/browse/RAD-663
https://jira.int.lmc.cz/browse/RAD-662

### DIFF
--- a/src/Http/Plugin/ExceptionPlugin.php
+++ b/src/Http/Plugin/ExceptionPlugin.php
@@ -30,7 +30,7 @@ final class ExceptionPlugin implements Plugin
             throw AuthorizationException::fromRequestAndResponse($request, $response);
         }
 
-        if ($responseCode >= 401 && $responseCode < 600) {
+        if ($responseCode >= 400 && $responseCode < 600) {
             // TODO: use more specific exceptions
             throw new RequestException($response->getReasonPhrase(), $request, $response);
         }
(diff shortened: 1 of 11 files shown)


kubasimon (approved):

src/Http/Plugin/ExceptionPlugin.php
```diff
@@ -30,7 +30,7 @@ private function transformResponseToException(RequestInterface $request, Respons
             throw AuthorizationException::fromRequestAndResponse($request, $response);
         }
 
-        if ($responseCode >= 401 && $responseCode < 600) {
+        if ($responseCode >= 400 && $responseCode < 600) {
```
davidkmenta, line 33: Any reason why not use the `StatusCodeInterface:: STATUS_BAD_REQUEST`?

OndraM: IMHO it would actually lower the readability.

The condition is now explicitly read as "for all response codes between 400 and 600 do this".
Using the constant, it will look like:
"for all response codes between BAD_REQUEST and 600 do this". But this is not what the condition is intended for.

The constants IMO makes sense when comparing with specific response code, not with range of codes...

davidkmenta (approved):

PR #19, opened 2017-11-22
Send client version header

OndraM:


jirinovak (dismissed):

davidkmenta (dismissed):

tests/Http/RequestManagerTest.php
```diff
@@ -53,5 +53,9 @@ public function shouldSendAndDecodeRequest(): void
             $recordedRequests[0]->getBody()->__toString()
         );
         $this->assertSame(['application/json'], $recordedRequests[0]->getHeader('Content-Type'));
+        $this->assertStringStartsWith(
+            'php-client/',
```
ghost: Why not assert the entire header value?

OndraM: Version will change every release. But ok, I can use the constant there, why not 🤷‍.

ghost (dismissed):

PR #20, opened 2017-11-22
Create and send requests through Matej instance

OndraM:
Final piece to actually interact with Matej API through the main `Matej` object instance 🎉 .

What is possible so far:

```php
$matej = new Matej('foo', 'apikey');

// Create new item property in database:
$response = $matej->request()
    ->setupItemProperties()
    ->addProperty(ItemPropertySetup::timestamp('valid_to'))
    ->send();

// Delete item property from database:
$response = $matej->request()
    ->deleteItemProperties()
    ->addProperty(ItemPropertySetup::timestamp('valid_from'))
    ->send();

// Send item property data to database
$response = $matej->request()
    ->events()
    ->addItemProperty(ItemProperty::create('1337', ['valid_from' => time(), 'title' => 'Title']))
    ->send();

```

Other commands and builders will follow (or will have an issue for grab).

### DIFF
--- a/src/Matej.php
+++ b/src/Matej.php
@@ -2,19 +2,61 @@
 
 namespace Lmc\Matej;
 
+use Http\Client\HttpClient;
+use Http\Message\MessageFactory;
+use Lmc\Matej\Http\RequestManager;
+use Lmc\Matej\Http\ResponseDecoderInterface;
+use Lmc\Matej\RequestBuilder\RequestBuilderFactory;
+
 class Matej
 {
     public const CLIENT_ID = 'php-client';
     public const VERSION = '0.0.0';
 
     /** @var string */
-    private $clientId;
+    private $accountId;
     /** @var string */
     private $apiKey;
+    /** @var RequestManager */
+    private $requestManager;
 
-    public function __construct(string $clientId, string $apiKey)
+    public function __construct(string $accountId, string $apiKey)
     {
-        $this->clientId = $clientId;
+        $this->accountId = $accountId;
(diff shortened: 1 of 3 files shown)


tests/MatejTest.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /** @test */
+    public function shouldBeInstantiable(): void
+    {
+        $matej = new Matej('accountId', 'apiKey');
+        $this->assertInstanceOf(Matej::class, $matej);
+    }
+
+    /** @test */
+    public function shouldExecuteRequestViaBuilder(): void
```
OndraM, line 22: Ideally, I would like to have an integration test there, which will interact with the real API. However this is tricky right now (we don't have available "unused" Matej test instance right now), so lets keep this as an isolated integration test...

src/Matej.php
```diff
@@ -2,19 +2,61 @@
...
     public const CLIENT_ID = 'php-client';
     public const VERSION = '0.0.0';
 
     /** @var string */
-    private $clientId;
+    private $accountId;
     /** @var string */
     private $apiKey;
+    /** @var RequestManager */
+    protected $requestManager;
```
jirinovak: Why `protected`?

We already have `protected` getter so subclass have acces to `$requestManager`.

OndraM: True, fixed.

foglcz (dismissed):

jirinovak (dismissed):

florianjiri (dismissed):

ghost (dismissed):

PR #26, opened 2017-11-23
Add UserMerge command

### LINKED ISSUE #23: User-merge command
- `Lmc\Matej\Model\Command\UserMerge`
- Static named constructor with helpful name (so you are sure which user is source and which target). Something like `UserMerge::toTargetUserMergeSourceUser($targetUser, $sourceUser)` :-).

Once the command is done, in separate commit (or in separate pull request) the EventsRequestBuilder should be extended by `addUserMerge` and `addUserMerges` methods.

foglcz:
First half of #23

This does not implement any actual call to Matej endpoints - those will be implemented in separate PRs.

### DIFF
--- a/src/Model/Command/UserMerge.php
+++ b/src/Model/Command/UserMerge.php
@@ -1,0 +1,50 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+/**
+ * Take all interactions from the source user and merge them to the target user.
+ * Source user will be DELETED and unknown to Matej from this action.
+ */
+class UserMerge extends AbstractCommand
+{
+    /** @var string */
+    private $sourceUserId;
+    /** @var string */
+    private $targetUserId;
+
+    private function __construct(string $targetUserId, string $sourceUserId)
+    {
+        $this->targetUserId = $targetUserId;
+        $this->sourceUserId = $sourceUserId;
+    }
+
+    /**
+     * Merge source user into target user AND DELETE SOURCE USER.
+     */
+    public static function mergeInto(string $targetUserId, string $sourceUserIdToBeDeleted): self
+    {
+        return new static($targetUserId, $sourceUserIdToBeDeleted);
+    }
+
+    /**
+     * Merge source user into target user AND DELETE SOURCE USER.
+     */
+    public static function mergeFromSourceToTargetUser(string $sourceUserIdToBeDeleted, string $targetUserId): self
+    {
+        return new static($targetUserId, $sourceUserIdToBeDeleted);
+    }
+
+    public function getCommandType(): string
+    {
+        return 'user-merge';
+    }
+
+    public function getCommandParameters(): array
+    {
+        return [
+            'target_user_id' => $this->targetUserId,
+            'source_user_id' => $this->sourceUserId,
+        ];
+    }
+}
--- a/tests/Model/Command/UserMergeTest.php
+++ b/tests/Model/Command/UserMergeTest.php
@@ -1,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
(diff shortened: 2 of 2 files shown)


src/Model/Command/UserMerge.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+class UserMerge extends AbstractCommand
+{
+    /** @var string */
+    private $sourceUserId;
+    /** @var string */
+    private $targetUserId;
+
+    /**
+     * Merge source user to target user AND DELETE SOURCE USER.
```
jirinovak: Same as in #27. I think we should merge constructor and class comments in this particular case.

src/Model/Command/UserMerge.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public function __construct(string $sourceUserId, string $targetUserId)
```
OndraM: All other Commands (ok, both of them :-D) could be instantiated only through the static constructor. My idea was to not mix two approaches (new vs static constructor) and even though some commands do need just one way of instantiation (thus the `::create()` method), I would prefer to make the static constructors the only way of Command instantiation (and make the `__construct()` private), so there is less cognitive load for the developer and also the usage is unified.

src/Model/Command/UserMerge.php
```diff
@@ -0,0 +1,52 @@
...
+        return [
+            'target_user_id' => $this->targetUserId,
+            'source_user_id' => $this->sourceUserId,
+        ];
+    }
+
+    /**
+     * Merge source user to target user AND DELETE SOURCE USER.
+     */
+    public static function create(string $sourceUserId, string $targetUserId): self
```
OndraM: I'm just thinking if having both of those constructors (which differs just with parameter order) will not make more mess... 🤔

OndraM: I tend to place the static named constructors right after __construct, so they are "together" on one place.

src/Model/Command/UserMerge.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        return new static($sourceUserId, $targetUserId);
+    }
+
+    /**
+     * Merge source user into target user AND DELETE SOURCE USER.
+     */
+    public static function mergeInto(string $targetUserId, string $sourceUserId): self
```
OndraM: Maybe make the "delete" thing more obvious by naming the variable eg. `$sourceUserIdToBeDeleted` :)?

tests/Model/Command/UserMergeTest.php
```diff
@@ -0,0 +1,45 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+use PHPUnit\Framework\TestCase;
+
+class UserMergeTest extends TestCase
+{
+    /**
+     * @test
```
OndraM: I prefer to use one-line annotations if it is possible.

OndraM (changes requested):

tests/Model/Command/UserMergeTest.php
```diff
@@ -0,0 +1,45 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+use PHPUnit\Framework\TestCase;
+
+class UserMergeTest extends TestCase
```
OndraM: We extend "our" TestCase in other tests... It has no added value here right now, but we may do it here as well to make it unified.

src/Model/Command/UserMerge.php
OndraM: So you'd like to have the PHPDoc as class comment or as constructor comment?

IMHO we may have some more detailed information and documentation in the class commend, and basic information in the constructor comment.

OndraM (dismissed):

jirinovak (dismissed):

Issue #24, opened 2017-11-23
Sorting command


OndraM:
- `Lmc\Matej\Model\Command\Sorting`
- Static named constructor with `string $userId` and `array $itemIds` params

PR #27, opened 2017-11-23
Add sorting command

foglcz:
First part of #24 

The actual call to an endpoint is not part of this pull request - it will be implemented once all `Matej\Model\Commands` are present, as the individual calls can interfere with each other (and we can implement only after #20 is merged)

### DIFF
--- a/src/Model/Command/Sorting.php
+++ b/src/Model/Command/Sorting.php
@@ -1,0 +1,42 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+/**
+ * Sorting items is a way how to use Matej to deliver personalized experience to users.
+ * It allows to sort given list of items according to the user preference.
+ */
+class Sorting extends AbstractCommand
+{
+    /** @var string */
+    private $userId;
+    /** @var string[] */
+    private $itemIds = [];
+
+    private function __construct(string $userId, array $itemIds)
+    {
+        $this->userId = $userId;
+        $this->itemIds = $itemIds;
+    }
+
+    /**
+     * Sort given item ids for user-based recommendations.
+     */
+    public static function create(string $userId, array $itemIds): self
+    {
+        return new static($userId, $itemIds);
+    }
+
+    public function getCommandType(): string
+    {
+        return 'sorting';
+    }
+
+    public function getCommandParameters(): array
+    {
+        return [
+            'user_id' => $this->userId,
+            'item_ids' => $this->itemIds,
+        ];
+    }
+}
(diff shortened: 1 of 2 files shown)


src/Model/Command/Sorting.php
```diff
@@ -0,0 +1,44 @@
...
+        return [
+            'user_id' => $this->userId,
+            'item_ids' => $this->itemIds,
+        ];
+    }
+
+    /**
+     * Sort given item ids for user-based recommendations
+     */
+    public static function create(string $userId, array $itemIds): self
```
jirinovak: Add another named constructor `sortFor` or so?

Like `mergeInto` from #26 .

src/Model/Command/Sorting.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+class Sorting extends AbstractCommand
+{
+    /** @var string */
+    private $userId;
+    /** @var string[] */
+    private $itemIds = [];
+
+    /**
+     * Sort given item ids for user-based recommendations
```
jirinovak: I think this is unnecessary as it is also in class comment. What do you think?

src/Model/Command/Sorting.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public function __construct(string $userId, array $itemIds)
+    {
+        $this->userId = $userId;
```
OndraM: In future, we may validate the values (and throw some exception on error) - like id length, format... I created an issue for this: https://github.com/lmc-eu/matej-client-php/issues/28

tests/Model/Command/SortingTest.php
```diff
@@ -0,0 +1,42 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+use PHPUnit\Framework\TestCase;
+
+class SortingTest extends TestCase
+{
+    /**
+     * @test
```
OndraM: dtto, could be one line

src/Model/Command/Sorting.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    /** @var string */
+    private $userId;
+    /** @var string[] */
+    private $itemIds = [];
+
+    /**
+     * Sort given item ids for user-based recommendations
+     */
+    public function __construct(string $userId, array $itemIds)
```
OndraM: What about making it also private :)?

tests/Model/Command/SortingTest.php
```diff
@@ -0,0 +1,42 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+use PHPUnit\Framework\TestCase;
+
+class SortingTest extends TestCase
```
OndraM: We extend "our" TestCase in other tests... It has no added value here right now, but we may do it here as well to make it unified.

OndraM (changes requested):

OndraM (dismissed):

jirinovak (dismissed):

PR #31, opened 2017-11-24
User recommendation command

OndraM:
~Not yet finished (+ with broken tests right now). However I'd like some feedback for the public API of the command.~

## Outdated issue description:
The most basic case would look like this:
```php
$response = $matej->request()
    ->recommendation(UserRecommendation::create('user-id', 10, 'scenario', 1.0, 3600, false))
    ->send();
```

But still the constructor has so many required params... Maybe at least the hard rotation could be optional (and default to false?). The other values must be explicitly defined and there is no "smart default", if I'm not mistaken, right?

And when you will need more customized recommendation, you will have to call setters on the UserRecommendation object:

 ```php
$recommendationCommand = UserRecommendation::create('user-id',10, 'scenario', 1.0, 3600, false);
$recommendationCommand->setFilter('item_id != 333');
$recommendationCommand->setMinimalRelevance(UserRecommendation::MINIMAL_RELEVANCE_HIGH);

$response = $matej->request()
    ->recommendation($recommendationCommand)
    ->send();
```

What do you think?

### DIFF
--- a/src/Model/Command/UserRecommendation.php
+++ b/src/Model/Command/UserRecommendation.php
@@ -1,0 +1,131 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+/**
+ * Deliver personalized recommendations for the given user.
+ */
+class UserRecommendation extends AbstractCommand
+{
+    public const MINIMAL_RELEVANCE_LOW = 'low';
+    public const MINIMAL_RELEVANCE_MEDIUM = 'medium';
+    public const MINIMAL_RELEVANCE_HIGH = 'high';
+
+    /** @var string */
+    protected $filterOperator = 'and';
+    /** @var string */
+    private $userId;
+    /** @var int */
+    private $count;
+    /** @var string */
+    private $scenario;
+    /** @var float */
+    private $rotationRate;
+    /** @var int */
+    private $rotationTime;
+    /** @var bool */
+    private $hardRotation = false;
+    /** @var string */
+    private $minimalRelevance = self::MINIMAL_RELEVANCE_LOW;
+    /** @var array */
+    private $filters = ['valid_to >= NOW'];
+
+    private function __construct(string $userId, int $count, string $scenario, float $rotationRate, int $rotationTime)
+    {
+        $this->userId = $userId; // TODO: assert format
+        $this->count = $count; // TODO: assert greater than 0
+        $this->scenario = $scenario; // TODO: assert format
+        $this->rotationRate = $rotationRate; // TODO: assert value between 0.0 and 1.0
+        $this->rotationTime = $rotationTime; // TODO: assert valid time interval
+    }
+
+    /**
+     * @param string $userId
+     * @param int $count Number of requested recommendations. The real number of recommended items could be lower or
+     * even zero when there are no items relevant for the user.
+     * @param string $scenario Name of the place where recommendations are applied - eg. 'search-results-page',
+     * 'emailing', 'empty-search-results, 'homepage', ...
+     * @param float $rotationRate How much should the item be penalized for being recommended again in the near future.
+     * Set from 0.0 for no rotation (same items will be recommended) up to 1.0 (same items should not be recommended).
+     * @param int $rotationTime Specify for how long will the item's rotationRate be taken in account and so the item
+     * is penalized for recommendations.
+     * @return UserRecommendation
+     */
+    public static function create(
+        string $userId,
+        int $count,
+        string $scenario,
+        float $rotationRate,
+        int $rotationTime
+    ): self {
+        return new static($userId, $count, $scenario, $rotationRate, $rotationTime);
+    }
+
+    /**
+     * Even with rotation rate 1.0 user could still obtain the same recommendations in some edge cases.
+     * To prevent this, enable hard rotation - recommended items are then excluded until rotation time is expired.
+     * By default hard rotation is not enabled.
+     */
+    public function enableHardRotation(): self
+    {
+        $this->hardRotation = true;
+
+        return $this;
+    }
+
+    /**
+     * Define threshold of how much relevant must the recommended items be to be returned.
+     * Default minimal relevance is "low".
+     */
+    public function setMinimalRelevance(string $minimalRelevance): self
+    {
+        // TODO: assert one of MIN_RELEVANCE_*
+        $this->minimalRelevance = $minimalRelevance;
+
+        return $this;
+    }
+
+    /**
+     * Add a filter to already added filters (including the default filter).
+     */
+    public function addFilter(string $filter): self
+    {
+        $this->filters[] = $filter;
+
+        return $this;
+    }
+
+    /**
+     * Overwrite all filters by custom one. Note this will override also the default filter.
+     */
+    public function setFilters(array $filters): self
+    {
+        $this->filters = $filters;
+
+        return $this;
+    }
+
+    protected function getCommandType(): string
+    {
+        return 'user-based-recommendations';
+    }
+
+    protected function getCommandParameters(): array
+    {
+        return [
+            'user_id' => $this->userId,
+            'count' => $this->count,
+            'scenario' => $this->scenario,
+            'rotation_rate' => $this->rotationRate,
+            'rotation_time' => $this->rotationTime,
+            'hard_rotation' => $this->hardRotation,
+            'min_relevance' => $this->minimalRelevance,
+            'filter' => $this->assembleFiltersString(),
+        ];
+    }
+
+    protected function assembleFiltersString(): string
+    {
+        return implode(' ' . $this->filterOperator . ' ', $this->filters);
+    }
+}
--- a/tests/Model/Command/UserRecommendationTest.php
+++ b/tests/Model/Command/UserRecommendationTest.php
@@ -1,0 +1,89 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model\Command;
+
+use PHPUnit\Framework\TestCase;
+
+class UserRecommendationTest extends TestCase
+{
+    /** @test */
+    public function shouldBeInstantiableViaNamedConstructorWithDefaultValues(): void
+    {
+        $command = UserRecommendation::create('user-id', 333, 'test-scenario', 1.0, 3600);
+
+        $this->assertInstanceOf(UserRecommendation::class, $command);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by OndraM** [comment]
Also the filters could be defined as an array of strings, and concatenated to output string only on build(). This will allow simpler manipulation with the default filter, adding custom filters atd.

Or could the default filter be defined as standalone property - what would allow to simply overwrite it eg. through RequestManager.

File: src/Model/Command/UserRecommendation.php [hunk 1]
```diff
@@ -0,0 +1,111 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Define threshold of how much relevant must the recommended items be to be returned.
+     */
+    public function setMinimalRelevance(string $minimalRelevance): self
+    {
+        // TODO: assert one of MIN_RELEVANCE_*
```
jirinovak: Maybe 3 setter should be provided e.g. `setLowMinimalRelevance` etc. and no validation will be needed.

src/Model/Command/UserRecommendation.php
```diff
@@ -0,0 +1,111 @@
...
+    /** @var float */
+    private $rotationRate;
+    /** @var int */
+    private $rotationTime;
+    /** @var bool */
+    private $hardRotation;
+    /** @var string */
+    private $minimalRelevance = self::MINIMAL_RELEVANCE_LOW;
+    /** @var string */
+    private $filter = 'valid_to >= NOW';
```
jirinovak: Isn`t this client specific?

`valid_to` must be present in item properties.

Or I am wrong?

jirinovak:
`The other values must be explicitly defined and there is no "smart default", if I'm not mistaken, right?`

Maybe smart default should also have `rotationRate` and `rotationTime`. Those default values may be provided by Matěj crew.


`Also the filters could be defined as an array of strings...`

I think filters as an array would be better.


`Or could the default filter be defined as standalone property - what would allow to simply overwrite it eg. through RequestManager.`

I don`t think we could find default filter that fits all clients. See my comment in review.

OndraM:
Yes, I agree it is, but having this item property in database and this
default filter is considered as best practice by the RaD team...  And the
API should follow the best practices (but allows you to explicitly change
that).

However, this should (in future) be at least somehow stated in the docs
more obviously...

Dne 24. 11. 2017 11:30 odp. napsal uživatel "Jiří Novák" <
notifications@github.com>:

src/Model/Command/UserRecommendation.php
kubasimon: Yes, its client specific.

Another argument is that 'filters' are client specific too (afaik all of them are still hardcoded).

So I suggests no default filter.

OndraM: Yes, it is client specific, but:
- all but one client ;-) will use it as for now, so we can make life easier for the vast majority  by defining the default filter like this;
- the one who does not use this default filter (but custome one), will anyway have to define the  filter -  so there will be no difference for him (define custom filter vs. overwrite default filter will be the same amount of work).

From the beginning we planned to have a simple way how to overwrite the default filter for all recommendations and sorting - so there will be no extra work for those not using the default filter. But for those who do use the default we will make the library adoption simpler, IMO.

Also, quoting Matej official docs:
> Matej supports following filter: `valid_to >= NOW` which should be used in each recommendation request.

I will talk about this issue with RAD team and we may change this behavior in the future, however, having this feature in the API client was partially their idea ¯\\\_(ツ)_/¯.

OndraM:
@jirinovak And what would you suggest as default value for rotation rate and rotation time? 

The main criteria whether this should be required in constructor or customisable via setter is the the ratio of "creating request with the default values" vs "creating request with custom values". If you - for example - use in 90 % of cases the defaults, lets do it via setters. But if the ratio is 50:50 or so, maybe the cunstructor params is a better and more transparent way.

src/Model/Command/UserRecommendation.php
jirinovak: There will be extra work for those who do not use default filter at all. They have to reset it. Or they end up with invalid request i guess.

It still seem weird to me to have default filter in such general library.

jirinovak:
As I said maybe the RAD team could provide `smart defaults`.

OndraM:
So cc @foglcz - any suggestions :)?

src/Model/Command/UserRecommendation.php
OndraM: No, they will not have to explicitly *reset* it. They will have to set custom filter anyway (and this will override the default one - if any). So no extra work for them. Because AFAIK there is no meaningful usecase for requesting recommendations without some kind of similar default filter.

foglcz: I agree. It might be useful though to have a way how to set the *default* filter for the customers use-case.

How about having a static variable that'll be the default value?

```php
UserRecommendation::$DEFAULTS['filter'] = null;
```

How do other libs solve this?

src/Model/Command/UserRecommendation.php
OndraM: Maybe... If you find this interesting, we may add it later (or send a PR :) ), it would be quite easy to implement. However we are a bit on time pressure for MVP :).

OndraM:
Ok, so:
- the default constructor requires just userId, recommendation count and scenario
- default value is used for all other properties with setters available for them
- I can imagine we may add in future eg. second contructor which will allow setting all properties as part of the instantiation (`::createConfigured($userId, $count, $scenario, $rotationRate, $rotationTime, $hardRotation, $minimalRelevance, $filters` - if someone will find this useful. 
- If you won't use the default filter, you will simple call `setFilters([...])` to overwrite it. Some simple way to set default filters for all Recommendations and Sortings may be added in the future.

### Basic usage with default values
```php
$response = $matej->request()
    ->recommendation(UserRecommendation::create('user-id', 10, 'scenario'))
    ->send();
```
=> this will use the default values:
- rotation_rate' => 1.0,
- 'rotation_time' => 3600,
- 'hard_rotation' => false,
- 'min_relevance' => 'low',
- 'filter' => 'valid_to >= NOW',

### Advanced usage with custom values
```php
$recommendationCommand = UserRecommendation::create('user-id', 10, 'scenario');
$recommendationCommand->setRotationRate(0.1337)
    ->setRotationTime(666)
    ->setMinimalRelevance(UserRecommendation::MINIMAL_RELEVANCE_HIGH)
    ->enableHardRotation()
    ->setFilters(['foo = bar', 'AND baz = ban']);

$response = $matej->request()
    ->recommendation($recommendationCommand)
    ->send();
```

jirinovak (dismissed):

kubasimon:
I'm missing 'allowNonexistent' parameter - not part of this pull request?

kubasimon:
And with current "smart defaults" is client able to omit some parameters (e.g. rotationRate/rotationTime etc) with $command->setRotationTime(null)? 

It's sometimes useful for some A/B tests on Matej side.

foglcz:
Got up to date feedback from RAD

* Rotation rate & time - request specific - should be part of constructor (ie. 2 weeks on mailing, 1 hour on serp)
* Minimal relevance - let's keep it as `MINIMAL_RELEVANCE_LOW` as default, as that will always return *something*. Some clients might want to use `MEDIUM` as min relevance, would add it as optional parameter of `UserRecommendation::create`
* Hard rotation & filter - can be as spec'ed

foglcz:
For filters, I'd personally refrain from using `->addFilter('AND some_filter')` notation, as that's a bit counter intuitive.

Can we keep the array, but do a `$filter = join($this->filterSeparator, $this->filters)` where `$this->filterSeparator = 'AND' && $this->filters = ['foo = bar', 'baz = ban']` ?

So that dev can use:

```php
$recommendationCommand = UserRecommendation::create('user-id', 10, 'scenario')
    ->setFilters(['foo = bar', 'baz = ban'])
    ->addFilter('bar = none')
;
```

OndraM:
@tenerd `allowNonexistent` is not part of API v2. I'm not sure what is recommended way of handling this case in the API v2 - but I guess RAD team will advice if you write them.

Also - all parameters are required in the API request according to JSON schema. And eg. the value of rotation rate must be like this: `"type_rotation_rate": { "type": "number", "minimum": 0, "maximum": 1.0 },`. So you cannot omit the param, nor set its value to `null`.

OndraM:
Ok, I updated the PR - rotationRate and rotationTime are now required in the constructor, without any default value.

Also the filters are now concatenated using filterOperator. However I kept this as protected, so one can eg. do custom implementation, what about that?

foglcz (dismissed):

src/Model/Command/UserRecommendation.php
```diff
@@ -0,0 +1,131 @@
...
+        int $count,
+        string $scenario,
+        float $rotationRate,
+        int $rotationTime
+    ): self {
+        return new static($userId, $count, $scenario, $rotationRate, $rotationTime);
+    }
+
+    /**
+     * Even with rotation rate 1.0 user could still obtain the same recommendations in some * edge cases.
```
davidkmenta: The asterisk between the `some` and `edge` words has any reason?

src/Model/Command/UserRecommendation.php
```diff
@@ -0,0 +1,131 @@
...
+     * To prevent this, enable hard rotation - recommended items are then excluded until rotation time is expired.
+     * By default hard rotation is not enabled.
+     */
+    public function enableHardRotation(): self
```
ghost: Why not `setHardRotation(bool $hardRotation)`?

ghost (dismissed):

jirinovak (dismissed):

src/Model/Command/UserRecommendation.php
OndraM: Yes, it has! It serves as an example of copy paste error 🙃 .

I will fix this, thanks for noticing.

src/Model/Command/UserRecommendation.php
OndraM: It thought this is easier to read, don't you think?

ghost: Yes, but you can't change the value back to `false`.

OndraM: Why would you need this?

And even if someone would need this, we can rather do `disableHardRotation()`...

foglcz (dismissed):

PR #32, opened 2017-11-25
Add missing invalid commnad response status

jirinovak:
According to docs some endpoints could return `INVALID` status.


OndraM:
Thanks. BTW the ERROR status is for "unhandled" internal exception, while INVALID is when Matej detects the command as invalid. Both of those error types are counted into "number of failed commands".

PR #40, opened 2017-11-27
Add first integration test case for Item Properties setup

foglcz:
* Uses environment variables to instantiate matej's client
* allows changing base url for local development

### DIFF
--- a/src/Matej.php
+++ b/src/Matej.php
@@ -39,6 +39,14 @@ class Matej
         return $this;
     }
 
+    /** @internal used mainly in integration tests */
+    public function setBaseUrl($baseUrl): self
+    {
+        $this->getRequestManager()->setBaseUrl($baseUrl);
+
+        return $this;
+    }
+
     /** @codeCoverageIgnore */
     public function setHttpMessageFactory(MessageFactory $messageFactory): self
     {
--- a/tests/IntegrationTests/IntegrationTestCase.php
+++ b/tests/IntegrationTests/IntegrationTestCase.php
@@ -1,0 +1,32 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\IntegrationTests;
+
+use Lmc\Matej\Matej;
+use Lmc\Matej\TestCase;
+
+class IntegrationTestCase extends TestCase
+{
+    /** @before */
+    protected function checkIfConfigured(): void
+    {
+        if (!getenv('MATEJ_TEST_ACCOUNTID')) {
+            $this->markTestSkipped('Environment variable MATEJ_TEST_ACCOUNTID has to be defined');
+        }
+
+        if (!getenv('MATEJ_TEST_APIKEY')) {
+            $this->markTestSkipped('Environment variable MATEJ_TEST_APIKEY has to be defined');
+        }
+    }
+
+    protected function createMatejInstance(): Matej
+    {
+        $instance = new Matej(getenv('MATEJ_TEST_ACCOUNTID'), getenv('MATEJ_TEST_APIKEY'));
+
+        if ($baseUrl = getenv('MATEJ_TEST_BASE_URL')) { // intentional assignment
+            $instance->setBaseUrl($baseUrl);
+        }
+
+        return $instance;
+    }
+}
--- a/tests/IntegrationTests/ItemPropertiesSetupTest.php
+++ b/tests/IntegrationTests/ItemPropertiesSetupTest.php
@@ -1,0 +1,122 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\IntegrationTests;
+
+use Lmc\Matej\Model\Command;
+
+/**
+ * @covers \Lmc\Matej\RequestBuilder\ItemPropertiesSetupRequestBuilder
+ * @covers \Lmc\Matej\RequestBuilder\EventsRequestBuilder
+ */
+class ItemPropertiesSetupTest extends IntegrationTestCase
+{
+    /**
+     * In Matej's Mongo Worker, this will throw an error, because the property isn't defined.
+     * The API will however return OK. This is desired behaviour.
+     *
+     * @test
+     */
+    public function shouldExecuteRequestContainingUnknownPropertyButPass(): void
+    {
+        $response = $this->createMatejInstance()
+            ->request()
+            ->events()
+            ->addItemProperty(
+                Command\ItemProperty::create(
+                    'integration-test-php-clientitem-id',
+                    ['integration_test_php_client_property_1' => true, 'integration_test_php_client_property_2' => false]
+                )
+            )
+            ->send();
+
+        $this->assertSame(1, $response->getNumberOfCommands());
+        $this->assertSame(1, $response->getNumberOfSuccessfulCommands());
+        $this->assertSame(0, $response->getNumberOfFailedCommands());
+        $this->assertSame(0, $response->getNumberOfSkippedCommands());
+
+        $commandResponse = $response->getCommandResponses()[0];
+        $this->assertSame('OK', $commandResponse->getStatus());
+        $this->assertSame([], $commandResponse->getData());
+        $this->assertSame('', $commandResponse->getMessage());
+    }
+
+    /**
+     * @test
+     * @depends shouldExecuteRequestContainingUnknownPropertyButPass
+     */
+    public function shouldCreateNewPropertiesInMatej(): void
+    {
+        $response = $this->createMatejInstance()
+            ->request()
+            ->setupItemProperties()
+            ->addProperty(Command\ItemPropertySetup::boolean('integration_test_php_client_bool'))
+            ->addProperty(Command\ItemPropertySetup::double('integration_test_php_client_double'))
+            ->addProperty(Command\ItemPropertySetup::int('integration_test_php_client_int'))
+            ->addProperty(Command\ItemPropertySetup::string('integration_test_php_client_string'))
+            ->addProperties([
+                Command\ItemPropertySetup::timestamp('integration_test_php_client_timestamp'),
+                Command\ItemPropertySetup::set('integration_test_php_client_set'),
+            ])
+            ->send();
+
+        $this->assertSame(6, $response->getNumberOfCommands());
+        $this->assertSame(6, $response->getNumberOfSuccessfulCommands());
+        $this->assertSame(0, $response->getNumberOfFailedCommands());
+        $this->assertSame(0, $response->getNumberOfSkippedCommands());
+    }
+
+    /**
+     * @test
+     * @depends shouldCreateNewPropertiesInMatej
+     */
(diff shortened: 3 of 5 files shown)


tests/IntegrationTests/IntegrationTestCase.php
```diff
@@ -0,0 +1,32 @@
...
+        if (!getenv('MATEJ_TEST_APIKEY')) {
+            $this->markTestSkipped('Environment variable MATEJ_TEST_APIKEY has to be defined');
+        }
+    }
+
+    protected function createMatejInstance(): Matej\Matej
+    {
+        $instance = new Matej\Matej(getenv('MATEJ_TEST_ACCOUNTID'), getenv('MATEJ_TEST_APIKEY'));
+
+        if ($baseUrl = getenv('MATEJ_TEST_BASE_URL')) { // intentional assignment
```
davidkmenta: IMHO such comments are useless.

src/Matej.php
```diff
@@ -55,7 +55,7 @@ public function setHttpResponseDecoder(ResponseDecoderInterface $responseDecoder
         return $this;
     }
 
-    protected function getRequestManager(): RequestManager
+    public function getRequestManager(): RequestManager
```
OndraM: I'd prefer to keep this as protected. The request manager is internal implementation detail, you should not know about it at all, when you interact with `Matej` object - this will also break the encapsulation.

Maybe we can add  `setBaseUrl()` method to `Matej` object instead?

tests/IntegrationTests/IntegrationTestCase.php
```diff
@@ -0,0 +1,32 @@
...
+        if (!getenv('MATEJ_TEST_ACCOUNTID')) {
+            $this->markTestSkipped('Environment variable MATEJ_TEST_ACCOUNTID has to be defined');
+        }
+
+        if (!getenv('MATEJ_TEST_APIKEY')) {
+            $this->markTestSkipped('Environment variable MATEJ_TEST_APIKEY has to be defined');
+        }
+    }
+
+    protected function createMatejInstance(): Matej\Matej
```
davidkmenta: Any reasons why not to import every class separately?

```
use Lmc\Matej\Matej;
use Lmc\Matej\Matej\TestCase;
```

etc...

tests/IntegrationTests/IntegrationTestCase.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Lmc\Matej\IntegrationTests;
+
+use Lmc\Matej;
+
+class IntegrationTestCase extends Matej\TestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
```
OndraM: Maybe use eg. `init()` method with `/** @before */` annotation, to avoid calling the parent::setUp?

tests/IntegrationTests/IntegrationTestCase.php
```diff
@@ -0,0 +1,32 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\IntegrationTests;
+
+use Lmc\Matej;
+
+class IntegrationTestCase extends Matej\TestCase
```
OndraM: We usually import whole namespaces.

tests/IntegrationTests/IntegrationTestCase.php
OndraM: Dtto - import whole namespace.

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+use Lmc\Matej\Model\Command;
+
+/**
+ * @covers ItemPropertiesSetupRequestBuilder
+ * @covers EventsRequestBuilder
+ */
+class ItemPropertiesSetupTest extends IntegrationTestCase
+{
+    /** @test This throws an error in Matej's MGW, but API will return OK */
```
OndraM: In case of description, `@depends` etc. I'd use multiline phpDoc :). Also in the other cases in this PR.

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -0,0 +1,104 @@
...
+    public function issueEventWithUnsetProperties(): void
+    {
+        $response = $this->createMatejInstance()->request()
+            ->events()
+            ->addItemProperty(
+                Command\ItemProperty::create(
+                    'unittest-phpsdk-item-id',
```
OndraM: We're referring to the library as "PHP API client", not SDK - so it can be reflected here as well :).

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * @covers ItemPropertiesSetupRequestBuilder
+ * @covers EventsRequestBuilder
+ */
+class ItemPropertiesSetupTest extends IntegrationTestCase
+{
+    /** @test This throws an error in Matej's MGW, but API will return OK */
+    public function issueEventWithUnsetProperties(): void
+    {
+        $response = $this->createMatejInstance()->request()
```
OndraM: Having only one `->` on one line is IMHO improves clarity, what do you think?

```php
    ->request() 
     ->events()
...
```

davidkmenta (dismissed):

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -0,0 +1,104 @@
...
+use Lmc\Matej\Model\Command;
+
+/**
+ * @covers ItemPropertiesSetupRequestBuilder
+ * @covers EventsRequestBuilder
+ */
+class ItemPropertiesSetupTest extends IntegrationTestCase
+{
+    /** @test This throws an error in Matej's MGW, but API will return OK */
+    public function issueEventWithUnsetProperties(): void
```
OndraM: We use BDD-like naming of the test methods, with "should" as the first part of the test name, to make clear what behavior the object under test *should* have.

ItemPropertiesSetup
... should[ExpectedBehaviorOfTheObjectUnderTest]
... shouldCreateNewItemProperties
... shouldDeleteItemProperties
... etc

OndraM (changes requested):

foglcz:
Updated as per review requests @davidkmenta @OndraM 

The only thing I left there was the `// intentional assignment` comment. I write them as a habit (similar to `// intentional ==`), so that it's immediately obvious to anyone exploring the codebase.

Happy to remove it, if general consensus is to not write them. @OndraM, what's your take?

src/Matej.php
```diff
@@ -39,6 +39,13 @@ public function setHttpClient(HttpClient $client): self
         return $this;
     }
 
+    public function setBaseUrl($baseUrl): self
```
OndraM: I'm thinking about either marking this as `@internal` or adding phpdoc, explaining the usecase is mostly for integration testing... But I don't have a strong feeling about it :)

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -0,0 +1,122 @@
...
+ */
+class ItemPropertiesSetupTest extends IntegrationTestCase
+{
+    /**
+     * In Matej's component MGW, this will throw an error, because the property isn't defined.
+     * The API will however return OK. This is desired behaviour.
+     *
+     * @test
+     */
+    public function shouldIssueEventWithUnknownPropertyButPassOk(): void
```
OndraM: shouldExecuteRequestContainingUnknownPropertyButPass :)?

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -0,0 +1,122 @@
...
+use Lmc\Matej\Model\Command;
+
+/**
+ * @covers ItemPropertiesSetupRequestBuilder
+ * @covers EventsRequestBuilder
+ */
+class ItemPropertiesSetupTest extends IntegrationTestCase
+{
+    /**
+     * In Matej's component MGW, this will throw an error, because the property isn't defined.
```
OndraM: MGW? 🤔

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -0,0 +1,122 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\IntegrationTests;
+
+use Lmc\Matej\Model\Command;
+
+/**
+ * @covers ItemPropertiesSetupRequestBuilder
+ * @covers EventsRequestBuilder
```
OndraM: We use FQCN in other tests

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -0,0 +1,122 @@
...
+            ->addProperties([
+                Command\ItemPropertySetup::timestamp('integration_test_php_client_timestamp'),
+                Command\ItemPropertySetup::set('integration_test_php_client_set'),
+            ])
+            ->send();
+
+        $this->assertSame(6, $response->getNumberOfCommands());
+        $this->assertSame(6, $response->getNumberOfSuccessfulCommands());
+        $this->assertSame(0, $response->getNumberOfFailedCommands());
+        $this->assertSame(0, $response->getNumberOfSkippedCommands());
```
OndraM, line 65: Too bad there is no "get item properties" endpoint in Matej - it would allow to write end-to-end integration tests, where you check the actual result in the database 🤷‍ .

tests/IntegrationTests/IntegrationTestCase.php
```diff
@@ -0,0 +1,32 @@
...
+        if (!getenv('MATEJ_TEST_APIKEY')) {
+            $this->markTestSkipped('Environment variable MATEJ_TEST_APIKEY has to be defined');
+        }
+    }
+
+    protected function createMatejInstance(): Matej
+    {
+        $instance = new Matej(getenv('MATEJ_TEST_ACCOUNTID'), getenv('MATEJ_TEST_APIKEY'));
+
+        if ($baseUrl = getenv('MATEJ_TEST_BASE_URL')) { // intentional assignment
```
OndraM, line 26: I'm familiar with getenv behavior so I wouldn't write it, but it does not bother me ¯\\\_(ツ)_/¯.

OndraM (changes requested):

tests/IntegrationTests/ItemPropertiesSetupTest.php
foglcz: RAD-720 created :)

OndraM:
Could you please fixup the commits before merging?

I see this whole as two separate commits: 
- Add ability to change base url (incl. tests)
- Add Item Properties setup integration test case

Everything else is just change of these commits in this PR, so there is no reason to have them as standalone commits.

OndraM (approved):

PR #41, opened 2017-11-27
Sorting request

### LINKED ISSUE #36: Sorting request builder
Usage will be like this:

```php
$response = $matej->request()
    ->sorting(Sorting::create(...))
    ->setInteraction(Interaction::purchase(...)) // optional
    ->setUserMerge(UserMerge::mergeInto('target', 'source')) // optional
    ->send();
```

foglcz:
~~PR currently references `feature/integration-test-case` until #40 is merged (to show nicer DIFF)~~

Closes #36 

Sorting command request. API usage:

```php
            ->request()
            ->sorting(Sorting::create('integration-test-php-client-user-id-A', ['itemA', 'itemB', 'itemC']))
            ->addUserMerge(UserMerge::mergeInto('user-id-A', 'user-id-B')) // optional
            ->addInteraction(Interaction::bookmark('user-id-A', 'itemA')) // optional
            ->send();
```

### DIFF
--- a/src/RequestBuilder/RequestBuilderFactory.php
+++ b/src/RequestBuilder/RequestBuilderFactory.php
@@ -3,6 +3,7 @@
 namespace Lmc\Matej\RequestBuilder;
 
 use Lmc\Matej\Http\RequestManager;
+use Lmc\Matej\Model\Command\Sorting;
 
 /**
  * Factory to create concrete RequestBuilder which helps you to create request for each Matej API
@@ -52,6 +53,14 @@ class RequestBuilderFactory
     public function campaign(): CampaignRequestBuilder
     {
         return $this->createConfiguredBuilder(CampaignRequestBuilder::class);
+    }
+
+    /**
+     * @return SortingRequestBuilder
+     */
+    public function sorting(Sorting $sorting): SortingRequestBuilder
+    {
+        return $this->createConfiguredBuilder(SortingRequestBuilder::class, $sorting);
     }
 
     // TODO: builders for other endpoints
--- a/src/RequestBuilder/SortingRequestBuilder.php
+++ b/src/RequestBuilder/SortingRequestBuilder.php
@@ -1,0 +1,49 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\RequestBuilder;
+
+use Fig\Http\Message\RequestMethodInterface;
+use Lmc\Matej\Model\Command\Interaction;
+use Lmc\Matej\Model\Command\Sorting;
+use Lmc\Matej\Model\Command\UserMerge;
+use Lmc\Matej\Model\Request;
+
+class SortingRequestBuilder extends AbstractRequestBuilder
+{
+    protected const ENDPOINT_PATH = '/sorting';
+
+    /** @var Interaction|null */
+    private $interactionCommand;
+    /** @var UserMerge|null */
+    private $userMergeCommand;
+    /** @var Sorting */
+    private $sortingCommand;
+
+    public function __construct(Sorting $sortingCommand)
+    {
+        $this->sortingCommand = $sortingCommand;
+    }
+
+    public function addUserMerge(UserMerge $merge): self
+    {
+        $this->userMergeCommand = $merge;
+
+        return $this;
+    }
+
+    public function addInteraction(Interaction $interaction): self
+    {
+        $this->interactionCommand = $interaction;
+
+        return $this;
+    }
+
+    public function build(): Request
+    {
+        return new Request(
+            self::ENDPOINT_PATH,
+            RequestMethodInterface::METHOD_POST,
+            [$this->interactionCommand, $this->userMergeCommand, $this->sortingCommand]
+        );
+    }
+}
--- a/tests/IntegrationTests/IntegrationTestCase.php
+++ b/tests/IntegrationTests/IntegrationTestCase.php
@@ -3,6 +3,7 @@
 namespace Lmc\Matej\IntegrationTests;
 
 use Lmc\Matej\Matej;
+use Lmc\Matej\Model\Response;
 use Lmc\Matej\TestCase;
 
 class IntegrationTestCase extends TestCase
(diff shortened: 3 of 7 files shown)


src/RequestBuilder/RequestBuilderFactory.php
```diff
@@ -46,6 +46,15 @@ public function events(): EventsRequestBuilder
         return $this->createConfiguredBuilder(EventsRequestBuilder::class);
     }
 
+    /**
+     * @param string[] $itemIds
+     * @return SortingRequestBuilder
+     */
+    public function sorting(string $userId, array $itemIds): SortingRequestBuilder
```
OndraM: What about passing Sorting command instance to this method instead? As shown here: https://github.com/lmc-eu/matej-client-php/issues/36

This introduces another way of creating Sorting command... And its also unsystematic - all methods of request Builders accept Command instances, there is no other "shortcut" for direct instantiation...

src/RequestBuilder/SortingRequestBuilder.php
```diff
@@ -0,0 +1,49 @@
...
+use Lmc\Matej\Model\Command\Sorting;
+use Lmc\Matej\Model\Command\UserMerge;
+use Lmc\Matej\Model\Request;
+
+class SortingRequestBuilder extends AbstractRequestBuilder
+{
+    protected const ENDPOINT_PATH = '/sorting';
+
+    /** @var Interaction */
+    private $interactionCommand = null;
```
OndraM: Null is default, no need to explicitly define it as a value. Maybe rather in the phpdoc, to say its "legal" value.

src/RequestBuilder/SortingRequestBuilder.php
```diff
@@ -0,0 +1,49 @@
...
+    /** @var UserMerge */
+    private $userMergeCommand = null;
+    /** @var Sorting */
+    private $sortingCommand;
+
+    public function __construct(string $userId, array $itemIds)
+    {
+        $this->sortingCommand = Sorting::create($userId, $itemIds);
```
OndraM: As mentioned, I'd keep this out of the builder and just accept Sorting instances.

Also in the Recommendation builder which will follow, this would require copying all the 5 params, which are required for the Command constructor there...

tests/IntegrationTests/IntegrationTestCase.php
```diff
@@ -29,4 +30,17 @@ protected function createMatejInstance(): Matej
 
         return $instance;
     }
+
+    protected function assertResponseCommandStatuses(Response $response, ...$expectedCommandStatuses): void
+    {
+        $this->assertSame(count($expectedCommandStatuses), $response->getNumberOfCommands());
```
OndraM, line 36: assertCount should be used instead

tests/IntegrationTests/SortingRequestTest.php
```diff
@@ -0,0 +1,72 @@
...
+use Lmc\Matej\Model\Command\UserMerge;
+
+/**
+ * @covers ItemPropertiesSetupRequestBuilder
+ * @covers EventsRequestBuilder
+ */
+class SortingRequestTest extends IntegrationTestCase
+{
+    /** @test */
+    public function shouldIssueSortingRequestOnly(): void
```
OndraM: What about use `execute` instead of `issue`? Its a bit confusing and uncommon verb for me...

tests/RequestBuilder/RequestBuilderFactoryTest.php
```diff
@@ -57,10 +60,16 @@ public function provideBuilderMethods(): array
             $builder->addItemProperty(ItemProperty::create('item-id', []));
         };
 
+        $sortingInit = function (SortingRequestBuilder $builder): void {
+            $builder->addUserMerge(UserMerge::mergeInto('user-a', 'user-b'));
```
OndraM: This should be just "minimal" setup to make the internal validation (eg. at least one command in other cases) passes. As there is this kind of validation, this whole could be omitted.

tests/RequestBuilder/SortingRequestBuilderTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $request = $builder->build();
+
+        $this->assertInstanceOf(Request::class, $request);
+        $this->assertSame(RequestMethodInterface::METHOD_POST, $request->getMethod());
+        $this->assertSame('/sorting', $request->getPath());
+
+        $requestData = $request->getData();
+        $this->assertCount(3, $requestData);
+        $this->assertContains($interactionCommand, $requestData);
```
OndraM: We should check the exact command order here.

Ie. index 0 is interaction, index 1 user mege, index 2 is sorting command.

```php
$this->assertSame($interactionCommand, $requestData[0]); 
$this->assertSame($userMergeCommand, $requestData[1]); 
$this->assertSame($sortingCommand, $requestData[2]); 
```

Also the sorting command check below would then not be necessary, wouldn't it?

tests/RequestBuilder/SortingRequestBuilderTest.php
```diff
@@ -0,0 +1,76 @@
...
+        $requestManagerMock = $this->createMock(RequestManager::class);
+        $requestManagerMock->expects($this->once())
+            ->method('sendRequest')
+            ->with($this->isInstanceOf(Request::class))
+            ->willReturn(new Response(0, 0, 0, 0));
+
+        $builder = new SortingRequestBuilder('userId1', ['itemId1', 'itemId2']);
+        $builder->setRequestManager($requestManagerMock);
+
+        $builder->addInteraction(Interaction::detailView('userId1', 'itemId1'));
```
OndraM: Maybe this addInteraction and addUserMerge are not necessary for purposes of this unit test.

OndraM (changes requested):

tests/IntegrationTests/IntegrationTestCase.php
foglcz: But assertCount takes "expected" as integer, and compares it to array. Here we take array, count/intersect it to get the expected value, and compare it to whatever the response contains.

Looks counter-intuitive on the first hand, but has correct error reporting (`failed asserting that $response->numberOfCommands() is not expected "3"`)

OndraM: Oh, I see, my fault, sorry :)

tests/IntegrationTests/SortingRequestTest.php
```diff
@@ -0,0 +1,73 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\IntegrationTests;
+
+use Lmc\Matej\Model\Command\Interaction;
+use Lmc\Matej\Model\Command\Sorting;
+use Lmc\Matej\Model\Command\UserMerge;
+
+/**
+ * @covers ItemPropertiesSetupRequestBuilder
```
OndraM: FQCN :)

OndraM (changes requested):

Issue #43, opened 2017-11-27
Enable integration tests on Travis


OndraM:
Integration tests on Travis (#40) are currently being skipped.

We should enable them (ie. set api key and account id via environment variables) once test instance of Matej is available (see RAD-697).

OndraM:
This is already fixed on GH Actions.

PR #46, opened 2017-11-28
Add missing integration tests

foglcz:
* Add integration tests for Campaign endpoint
* Add integration tests for Events endpoint
* Add integration tests for rate limiting (cannot send more than 1000 commands in a batch)

**Note:** Rate limiting is not implemented on Matej's side and these tests fail until new version is released (being implemented as part of RAD-719)


tests/IntegrationTests/CampaignBuilderTest.php
```diff
@@ -0,0 +1,75 @@
...
+namespace Lmc\Matej\IntegrationTests;
+
+use Lmc\Matej\Exception\RequestException;
+use Lmc\Matej\Model\Command\Sorting;
+use Lmc\Matej\Model\Command\UserRecommendation;
+
+/**
+ * @covers \Lmc\Matej\RequestBuilder\CampaignRequestBuilder
+ */
+class CampaignBuilderTest extends IntegrationTestCase
```
jirinovak: Maybe `@group integration` could be added. For fine grained test execution.

tests/IntegrationTests/CampaignBuilderTest.php
```diff
@@ -0,0 +1,75 @@
...
+        $this->assertResponseCommandStatuses($response, 'OK');
+    }
+
+    /** @test */
+    public function shouldExecuteSortingCommandOnly(): void
+    {
+        $response = $this->createMatejInstance()
+            ->request()
+            ->campaign()
+            ->addSorting(Sorting::create('integration-test-php-client-user-id-A', ['itemA', 'itemB', 'itemC']))
```
jirinovak: Provide

jirinovak: Provide `createSortingCommand` cause same command is created multiple times.

tests/IntegrationTests/CampaignBuilderTest.php
```diff
@@ -0,0 +1,75 @@
...
+            ->campaign()
+            ->addRecommendation($this->createRecommendationCommand())
+            ->addSorting(Sorting::create('integration-test-php-client-user-id-A', ['itemA', 'itemB', 'itemC']))
+            ->send();
+
+        $this->assertResponseCommandStatuses($response, 'OK', 'OK');
+    }
+
+    /** @test */
+    public function shouldExecuteThreeThousandCommandsAndFail(): void
```
jirinovak: I wouldn't introduce failing test. Let's wait to RAD team implement rate limit and than add these tests.

tests/IntegrationTests/EventsBuilderTest.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        for ($i = 0; $i < $amount; $i++) {
+            $builder->addItemProperty(
+                ItemProperty::create('test-item-id' . $i, ['test_boolean' => (bool) ($i % 2)])
+            );
+        }
+    }
+
+    /** @return string[] */
+    private function generateOkStatuses(int $amount): array
```
jirinovak: This will be useful in other integration tests. Move to `IntegrationTestCase`?

tests/IntegrationTests/EventsBuilderTest.php
```diff
@@ -0,0 +1,156 @@
...
+        $builder = $this->createMatejInstance()->request()->events();
+        $this->appendInteractionCommands($builder);
+        $this->appendItemPropertyCommands($builder);
+        $response = $builder->send();
+
+        $this->assertResponseCommandStatuses($response, ...$this->generateOkStatuses(2));
+    }
+
+    /** @test */
+    public function shouldExecuteUserMergeAndItemPropertyCommandsOnly(): void
```
jirinovak: There are many command combinations. IMO we should test only these with real use case.

`[merge], [recommendation]`

jirinovak: There are many command combinations. IMO we should test only these with real use case.

`[merge], [recommendation]`

`[interaction], [recommendation]`

jirinovak: There are many command combinations. IMO we should test only these with real use case.

`[merge], [recommendation]`

`[interaction], [recommendation]`

And maybe some variant with sorting.

jirinovak (changes requested):

tests/IntegrationTests/CampaignBuilderTest.php
OndraM: I agree we need to control which tests are run. But Isuggest keeping this as-is for now - because before we do #43 I plan to setup phpunit.xml so that IntegrationTests directory will be in separate suite (`phpunit --suite integration`), so you don't need to define the test groups like this.

tests/IntegrationTests/CampaignBuilderTest.php
OndraM: Instead I'd rather throw exception right from the client when you attempt to build request with >1000 requests. So IMHO this test should be removed.

And in future we may also implement some chunking in the client.

tests/IntegrationTests/EventsBuilderTest.php
OndraM: +1, I don't think there is much value in testing all possible path combinations in integration tests... Also keep on mind they take time and generate actual traffic to Matej.

tests/IntegrationTests/EventsBuilderTest.php
```diff
@@ -0,0 +1,156 @@
...
+        $this->appendInteractionCommands($builder);
+        $this->appendUserMergeCommands($builder);
+        $this->appendItemPropertyCommands($builder);
+        $response = $builder->send();
+
+        $this->assertResponseCommandStatuses($response, ...$this->generateOkStatuses(3));
+    }
+
+    /** @test */
+    public function shouldExecuteOneThousandCommands(): void
```
OndraM: I wouldn't test this at all... This should be part of Matej tests, not part of integration tests for the PHP client - they should check the integration between the systems, not complete behavior of Matej.

tests/IntegrationTests/EventsBuilderTest.php
```diff
@@ -0,0 +1,156 @@
...
+        $this->appendInteractionCommands($builder, null, 332);
+        $this->appendUserMergeCommands($builder, 332);
+        $this->appendItemPropertyCommands($builder, 332);
+        $response = $builder->send();
+
+        $this->assertResponseCommandStatuses($response, ...$this->generateOkStatuses(1000));
+    }
+
+    /** @test */
+    public function shouldExecuteThreeThousandCommandsAndFail(): void
```
OndraM: dtto, imho not needed

tests/IntegrationTests/EventsBuilderTest.php
OndraM: I wouldn't test this here at all... This should be part of Matejs own test, not integration tests of PHP API client - we should check integration between the two systems, not overall Matej behavior....

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -119,4 +121,29 @@ public function shouldDeleteCreatedPropertiesFromMatej(): void
         $this->assertSame(0, $response->getNumberOfFailedCommands());
         $this->assertSame(0, $response->getNumberOfSkippedCommands());
     }
+
+    /**
+     * @dataProvider provideBuilders
+     * @test
```
OndraM: `@test` is placed before `@dataProvider` in other tests

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -119,4 +121,29 @@ public function shouldDeleteCreatedPropertiesFromMatej(): void
...
+     */
+    public function shouldExecuteThreeThousandSetupCommandsAndFail(ItemPropertiesSetupRequestBuilder $builder): void
```
OndraM: But again... I would not do this :)

OndraM (changes requested):

phpunit.xml.dist
```diff
@@ -6,7 +6,10 @@
 >
     <testsuites>
         <testsuite name="Unit tests">
```
OndraM: I suggest naming the suites "unit" and "integration" so it could be started more easily from the CLI:

`vendor/bin/phpunit --testsuite unit`
`vendor/bin/phpunit --testsuite integration`

OndraM (changes requested):

PR #47, opened 2017-11-29
Validate command data

### LINKED ISSUE #28: Validate command data
We should validate data passed to commands in corresponding setters (or constructors), so invalid data are not even send to Matej API (which would reject them).

Eg. user_id (and other ids) must conform (according to JSON schema) ` "type_identifier": { "type": "string", "pattern": "^[0-9A-Za-z-_]+$", "maxLength": 100 }` etc.

- Use assertion library (https://github.com/beberlei/assert or https://github.com/webmozart/assert)
- Throw custom exception
- Add information about the required value also to PHPDoc.

OndraM:
As mentioned in #28 , the plan is to allow setting only those values, allowed by Matej's API (and Matej JSON schema), so that you will get the feedback that your data are wrong sooner, even without sending them to Matej API (which will reject them).

I'd like to add the assertions in a similar way (via protected constructors - to not put all the assertions into constructor) to all Commands etc. Any comments, ideas?

### DIFF
--- a/src/Exception/DomainException.php
+++ b/src/Exception/DomainException.php
@@ -1,0 +1,47 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Exception;
+
+use Assert\AssertionFailedException;
+
+/**
+ * Exception thrown when invalid value is passed while creating domain model.
+ *
+ * @codeCoverageIgnore
+ */
+class DomainException extends LogicException implements AssertionFailedException
+{
+    /** @var string|null */
+    private $propertyPath;
+    /** @var mixed */
+    private $value;
+    /** @var array */
+    private $constraints;
+
+    public function __construct(string $message, int $code, ?string $propertyPath, $value, array $constraints = [])
+    {
+        parent::__construct($message, $code);
+
+        $this->propertyPath = $propertyPath;
+        $this->value = $value;
+        $this->constraints = $constraints;
+    }
+
+    public function getPropertyPath(): ?string
+    {
+        return $this->propertyPath;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    public function getConstraints(): array
+    {
+        return $this->constraints;
+    }
+}
--- a/src/Model/Assertion.php
+++ b/src/Model/Assertion.php
@@ -1,0 +1,30 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model;
+
+use Lmc\Matej\Exception\DomainException;
+
+/**
+ * @method static bool allTypeIdentifier(mixed $value) Assert value is valid Matej type identifier for all values
+ */
+class Assertion extends \Assert\Assertion
+{
+    protected static $exceptionClass = DomainException::class;
+
+    /**
+     * Assert value is valid Matej type identifier
+     * @param mixed $value
+     */
+    public static function typeIdentifier($value): bool
+    {
+        static::regex(
+            $value,
+            '/^[0-9A-Za-z-_]+$/',
+            'Value "%s" does not match type identifier format requirement (must contain only of alphanumeric chars,'
+            . ' dash or underscore)'
+        );
+        static::maxLength($value, 100);
+
+        return true;
+    }
+}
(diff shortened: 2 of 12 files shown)


src/Exception/DomainException.php
```diff
@@ -0,0 +1,45 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Exception;
+
+use Assert\AssertionFailedException;
+
+/**
+ * Exception thrown invalid value is passed while creating domain model.
```
MarketaSebkova: `Exception thrown when...`?

src/Model/Assertion.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Lmc\Matej\Exception\DomainException;
+
+class Assertion extends \Assert\Assertion
+{
+    protected static $exceptionClass = DomainException::class;
+
+    /**
+     * Assert value is valid Matej type identifier
+     * @param mixed $value
```
Fantomas-7: `@return`?

src/Model/Command/Interaction.php
```diff
@@ -34,12 +36,12 @@ private function __construct(
...
-        $this->itemId = $itemId; // TODO: assert format
-        $this->value = $value; // TODO: assert value between 0-1
-        $this->context = $context; // TODO: assert format
-        $this->timestamp = $timestamp ?: time(); // TODO: assert format
+        $this->setInteractionType($interactionType);
+        $this->setUserId($userId);
+        $this->setItemId($itemId);
+        $this->setValue($value);
+        $this->setContext($context);
+        $this->setTimestamp($timestamp ?: time());
```
ghost: Why not use `??` instead of `?:`

src/Model/Command/Interaction.php
```diff
@@ -121,4 +123,51 @@ public function getCommandParameters(): array
...
+    protected function setContext(string $context): void
+    {
+        Assertion::typeIdentifier($context);
+
+        $this->context = $context;
+    }
+
+    protected function setTimestamp(int $timestamp): void
+    {
+        Assertion::greaterThan($timestamp, 0);
```
ghost, line 157: Isn't `$timestamp = 0` technically valid?

ghost (changes requested):

MarketaSebkova (changes requested):

src/Model/Command/Interaction.php
OndraM: `??` is useful when the value may not be set. But the `$timestamp` is always set (but may be null).

src/Model/Command/Interaction.php
OndraM: Technically maybe yes, but not as per Matej JSON schema: `"type_timestamp": { "type": "integer", "minimum": 1 },`

src/Model/Command/Interaction.php
ghost: `??` is also useful when the value might be `null`, which is the default value here. With `?:` both `null` and `0` get coalesced to `time()`. I just wasn't sure if that was intended.

src/Model/Command/Interaction.php
ghost: ¯\\\_(ツ)\_/¯

src/Model/Command/Interaction.php
OndraM: I will change the condition to `$timestamp === null`  to clearly state this covers the case when no value is passed (ie. default value should be used).

src/Model/Command/ItemPropertySetup.php
```diff
@@ -58,6 +57,27 @@ public static function set(string $propertyName): self
...
+    protected function setPropertyName(string $propertyName): void
+    {
+        Assertion::typeIdentifier($propertyName);
+
+        $this->propertyName = $propertyName;
+    }
+
+    protected function setPropertyType(string $propertyType): void
+    {
+        Assertion::choice($propertyType, [
```
jirinovak: This is redundant, isn't it?

I cannot create invalid `ItemPropertySetup` cause private `constructor`.

And factory constructors are covered by tests.

davidkmenta (dismissed):

jirinovak (dismissed):

Issue #18, opened 2017-11-22
Add Matej-Request-Id header


OndraM:
Each request should contain `Matej-Request-Id` header with unique value. (maybe UUIDv4 via https://github.com/ramsey/uuid)

There should also be some documented way how to dump this ID for debugging purposes. 

Also response always contains `Matej-Response-Id` header, so it should be somehow accessible as well.

OndraM:
Waiting for implementation in Matej API.

PR #50, opened 2017-12-04
Allow to set custom request id and decode response id from the response

OndraM:
Useful for debugging concrete requests. See #18 and RAD-653.

Mark request with custom ID to make it traceable by R&D team:
```
$response = $matej->request()
    ->recommendation($recommendation)
    ->setRequestId('this-is-request-id-i-want-to-trace')
    ->send();
```

Or log response ID even if no custom was send with the request:

```
$response = $matej->request()
    ->recommendation($recommendation)
    ->send();

if (/* I don't like something in the response */) {
    $this->logger->error($response->getResponseId()); // there will be some random response ID, which you can ask R&D team to debug
}
```

### DIFF
--- a/src/Http/RequestManager.php
+++ b/src/Http/RequestManager.php
@@ -23,6 +23,7 @@ class RequestManager
 {
     public const CLIENT_VERSION_HEADER = 'Matej-Client-Version';
     public const REQUEST_ID_HEADER = 'Matej-Request-Id';
+    public const RESPONSE_ID_HEADER = 'Matej-Response-Id';
 
     /** @var string */
     private $baseUrl = 'https://%s.matej.lmc.cz';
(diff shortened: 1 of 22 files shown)


foglcz (approved):

jirinovak (approved):

src/Http/RequestManager.php
```diff
@@ -127,7 +129,10 @@ protected function createHttpRequestFromMatejRequest(Request $request): RequestI
             ->createRequest(
                 $request->getMethod(),
                 $uri,
-                ['Content-Type' => 'application/json'],
+                [
+                    'Content-Type' => 'application/json',
```
davidkmenta, line 133: ehmm... maybe you should use this awesome library https://github.com/lmc-eu/http-constants 😄

davidkmenta (approved):

src/Http/RequestManager.php
OndraM: @davidkmenta Yes, I heard about its awesomeness as well - but its IMO more useful when building larger app where you use those headers more frequently. As for the library, it won't pay for itself (for now) - and I'm trying to not expand the dependencies footprint of this library if not necessary.

PR #51, opened 2017-12-04
Add check for forbidden item_id in item properties (RAD-720)

foglcz:


src/Model/Command/ItemProperty.php
```diff
@@ -17,7 +17,13 @@ class ItemProperty extends AbstractCommand
     private function __construct(string $itemId, array $properties)
     {
         $this->setItemId($itemId);
+
         $this->properties = $properties;
+        Assertion::keyNotExists(
```
OndraM: I'd create protected setter for properties an put the assertion inside.

OndraM (changes requested):

OndraM:
And what is the reason for this? Its because the item property name is there by default always?

And the item_id cannot be updated once item is created?

Maybe the reasons should be also explained in the exception message.

foglcz:
+ add chagnelog

foglcz:
Right. So, `item_id` is something like a primary key. In original tests:

```php
$ip = ItemProperty::create('exampleItemId', ['item_id' => 'someOtherItemId']);
dump($ip->getCommandParameters());
// ['item_id' => 'examleItemId']
```

While correct behaviour (property should not overwrite the id of the item), I felt that this would be confusing to the developer.

tests/Model/Command/ItemPropertyTest.php
```diff
@@ -24,6 +25,15 @@ public function shouldBeInstantiableViaNamedConstructor(array $properties, array
         );
     }
 
+    /**
+     * @test
+     */
+    public function shouldNotAllowItemIdInProperties(): void
```
OndraM: This should be placed after the following data provider

src/Model/Command/ItemProperty.php
```diff
@@ -32,6 +32,17 @@ protected function setItemId(string $itemId): void
         $this->itemId = $itemId;
     }
 
+    protected function setProperties(array $properties): void
+    {
+        Assertion::keyNotExists(
+            $properties,
+            'item_id',
+            '$properties cannot contain key "item_id" - please use different key name.'
```
OndraM: I'm not sure "please use different key name" is the best advice.

Maybe we should rather explain the item_id is already used for identifying the item and it cannot be updated.

src/Model/Command/ItemPropertySetup.php
```diff
@@ -60,6 +60,7 @@ public static function set(string $propertyName): self
     protected function setPropertyName(string $propertyName): void
     {
         Assertion::typeIdentifier($propertyName);
+        Assertion::notEq($propertyName, 'item_id', 'Cannot define property with name item_id - please use different name');
```
OndraM: Dtto, maybe instead explain the item_id is set up in Matej by default and don't need to be set up explicitly.

OndraM (changes requested):

tests/Model/Command/ItemPropertyTest.php
```diff
@@ -2,10 +2,20 @@
...
 use PHPUnit\Framework\TestCase;
 
 class ItemPropertyTest extends TestCase
 {
+    /**
+     * @test
+     */
+    public function shouldNotAllowItemIdInProperties(): void
+    {
+        $this->expectException(DomainException::class);
```
OndraM, line 15: I'd add expectExceptionMessage - to make sure the exception occurred in the expected case.

src/Model/Command/ItemPropertySetup.php
```diff
@@ -60,6 +60,11 @@ public static function set(string $propertyName): self
     protected function setPropertyName(string $propertyName): void
     {
         Assertion::typeIdentifier($propertyName);
+        Assertion::notEq(
+            $propertyName,
+            'item_id',
+            'Cannot redefine property item_id - it is used as primary key to identify items. Please use different name.'
```
OndraM: `Cannot manipulate with property "item_id" - it is used by Matej to identify items.` or something?

src/Model/Command/ItemProperty.php
```diff
@@ -32,6 +32,17 @@ protected function setItemId(string $itemId): void
         $this->itemId = $itemId;
     }
 
+    protected function setProperties(array $properties): void
+    {
+        Assertion::keyNotExists(
+            $properties,
+            'item_id',
+            '$properties cannot contain key item_id - it is used as primary key to identify the item. Please use different key instead.'
```
OndraM: `Cannot update value of "item_id" property - it is used by Matej to identify the item and cannot be altered once created.` ?

OndraM (changes requested):

OndraM (approved):

PR #53, opened 2017-12-04
Add endpoint for getting item properties

foglcz:
Depends on matej's deploy of RAD-720

Depends on #46 - **before merging, change base of PR**

Implements `GET /item-properties`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 <!-- There is always Unreleased section on the top. Subsections (Added, Changed, Fixed, Removed) should be added as needed. -->
 
 ## Unreleased
+### Added
+- Endpoint to get all defined item properties in the Matej database (`$matej->request()->getItemProperties()`).
 
 ## 1.0.0 - 2017-12-05
 ### Added
--- a/src/RequestBuilder/ItemPropertiesGetRequestBuilder.php
+++ b/src/RequestBuilder/ItemPropertiesGetRequestBuilder.php
@@ -1,0 +1,16 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\RequestBuilder;
+
+use Fig\Http\Message\RequestMethodInterface;
+use Lmc\Matej\Model\Request;
+
+class ItemPropertiesGetRequestBuilder extends AbstractRequestBuilder
+{
+    protected const ENDPOINT_PATH = '/item-properties';
+
+    public function build(): Request
+    {
+        return new Request(self::ENDPOINT_PATH, RequestMethodInterface::METHOD_GET, [], $this->requestId);
+    }
+}
--- a/README.md
+++ b/README.md
@@ -91,6 +91,11 @@ $response = $matej->request()
     ->addProperty(ItemPropertySetup::string('title'))
     ->send();
 
+// Get list of item properties that are defined in matej
+$response = $matej->request()
+    ->getItemProperties()
+    ->send();
+
 // Delete item property from database:
 $response = $matej->request()
     ->deleteItemProperties()
--- a/src/Model/Request.php
+++ b/src/Model/Request.php
@@ -18,7 +18,7 @@ class Request
     /** @var string */
     private $requestId;
 
-    public function __construct(string $path, string $method, array $data, string $requestId = null)
+    public function __construct(string $path, string $method, array $data = [], string $requestId = null)
     {
         $this->path = $path;
         $this->method = $method;
--- a/src/RequestBuilder/RequestBuilderFactory.php
+++ b/src/RequestBuilder/RequestBuilderFactory.php
@@ -19,6 +19,11 @@ class RequestBuilderFactory
         $this->requestManager = $requestManager;
     }
 
+    public function getItemProperties(): ItemPropertiesGetRequestBuilder
+    {
+        return $this->createConfiguredBuilder(ItemPropertiesGetRequestBuilder::class);
+    }
+
     /**
      * Define new properties into the database. Those properties will be created and subsequently accepted by Matej.
      *
--- a/tests/integration/RequestBuilder/ItemPropertiesGetRequestBuilderTest.php
+++ b/tests/integration/RequestBuilder/ItemPropertiesGetRequestBuilderTest.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\IntegrationTests\RequestBuilder;
+
+use Lmc\Matej\IntegrationTests\IntegrationTestCase;
+
+/**
+ * @covers \Lmc\Matej\RequestBuilder\ItemPropertiesGetRequestBuilder
+ */
+class ItemPropertiesGetRequestBuilderTest extends IntegrationTestCase
+{
+    /** @test */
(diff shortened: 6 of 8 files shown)


src/RequestBuilder/ItemPropertiesModifierRequestBuilder.php
```diff
@@ -7,10 +7,8 @@
 use Lmc\Matej\Model\Command\ItemPropertySetup;
 use Lmc\Matej\Model\Request;
 
-class ItemPropertiesSetupRequestBuilder extends AbstractRequestBuilder
+class ItemPropertiesModifierRequestBuilder extends ItemPropertiesGetterRequestBuilder
```
OndraM: My preference is to keep this name as is - its meaning is already IMHO clear and distinctive from the GET version, and this adds no extra value - maybe even is confusing, because you actually cannot *modify* (like change data type) item properties once they are created in the database. So the the `ItemPropertiesSetupRequestBuilder` is for me ok. What do you think?

src/RequestBuilder/ItemPropertiesGetterRequestBuilder.php
```diff
@@ -0,0 +1,16 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\RequestBuilder;
+
+use Fig\Http\Message\RequestMethodInterface;
+use Lmc\Matej\Model\Request;
+
+class ItemPropertiesGetterRequestBuilder extends AbstractRequestBuilder
```
OndraM: Its a builder, which build request to get item properties (Its not building "getter request"), so why not just `ItemPropertiesGetRequestBuilder`? (But only if the other one is kept as `ItemPropertiesSetupRequestBuilder`)

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -101,22 +117,47 @@ public function shouldExecuteEventWithJustCreatedProperties(): void
...
+     * Fill the request builder with given properties
+     *
+     * @param string[][] $properties type => name
+     */
+    private function fillPropertyRequestBuilder(
+        ItemPropertiesModifierRequestBuilder $builder,
+        array $properties
+    ): ItemPropertiesModifierRequestBuilder {
+        foreach ($properties as $name => $definition) {
+            $property = forward_static_call([Command\ItemPropertySetup::class, $definition[0]], $name);
```
OndraM: `ItemPropertySetup::$definition[0]($name)` :)?

tests/IntegrationTests/ItemPropertiesSetupTest.php
```diff
@@ -101,22 +117,47 @@ public function shouldExecuteEventWithJustCreatedProperties(): void
...
+            $builder->addProperty($property);
+        }
+
+        return $builder;
+    }
+
+    /**
+     * @return array[string, array[string, mixed]]
```
OndraM: I'm not sure there is any common PHP standard of defining internal array structure (and PHPStorm don't understand this, no?) , so we may just stick with standardized `array[]`...

src/RequestBuilder/ItemPropertiesModifierRequestBuilder.php
OndraM: If the inheritance is there just because of the endpoint path, I'd not introduce it at all and rather have the endpoint path copied on two places. "Setup request builder" is from my POV not specific version of "Get request builder". It would only make sense IMHO to have `AbstractItemPropertiesRequestBuilder` with two childs, but would be a bit overengineered :).

OndraM (commented):
Could you also add some exmaple usage to readme? (But keep it short - as this is just an "auxiliary" endpoint used mainly for debugging, and not a main feature :) ).

Also something is missing code coverage according to coveralls, could you have a look, please?

src/RequestBuilder/ItemPropertiesGetRequestBuilder.php
```diff
@@ -0,0 +1,16 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\RequestBuilder;
+
+use Fig\Http\Message\RequestMethodInterface;
+use Lmc\Matej\Model\Request;
+
+class ItemPropertiesGetRequestBuilder extends AbstractRequestBuilder
```
OndraM, line 8: Could you please add also simple unit test for this builder? (See other unit tests for builders).

OndraM (changes requested):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 <!-- There is always Unreleased section on the top. Subsections (Added, Changed, Fixed, Removed) should be added as needed. -->
 
 ## Unreleased
+- Endpoint `$matej->getItemProperties()` that returns all defined item properties
```
OndraM:
```
### Added
- Endpoint to get all defined item properties in the Matej database (`$matej->request()->getItemProperties()`).
```

src/RequestBuilder/ItemPropertiesSetupRequestBuilder.php
```diff
@@ -7,10 +7,8 @@
 use Lmc\Matej\Model\Command\ItemPropertySetup;
 use Lmc\Matej\Model\Request;
 
-class ItemPropertiesSetupRequestBuilder extends AbstractRequestBuilder
+class ItemPropertiesSetupRequestBuilder extends ItemPropertiesGetRequestBuilder
```
OndraM: What do you think about this? (The [comment](https://github.com/lmc-eu/matej-client-php/pull/53#discussion_r154908592) got "outdated" by github after the rebase...)

> If the inheritance is there just because of the endpoint path, I'd not introduce it at all and rather have the endpoint path copied on two places. "Setup request builder" is from my POV not specific version of "Get request builder". It would only make sense IMHO to have `AbstractItemPropertiesRequestBuilder` with two childs, but would be a bit overengineered :).

foglcz: This is a very good point; I was thinking from the perspective of REST endpoints - and not from the perspective of RequestBuilders.

Will update code in a sec.

tests/IntegrationTests/ItemPropertiesSetupTest.php
foglcz: That wouldn't work as PHP would think that $definition is a static array on `ItemPropertySetup` class.

Would have to be `ItemPropertySetup::{$definition[0]}($name)` which I found pretty ugly :|

tests/IntegrationTests/ItemPropertiesSetupTest.php
foglcz: It's true that phpstorm doesn't understand this notation, but PHPStan does :)

OndraM: Is it documented somewhere in PHPStan? I couldn't find it...

OndraM (approved):

OndraM (changes requested):
If you could please rebase the PR (to get rid of the fixup commits) and we can merge it then 🎉

OndraM (approved):

PR #55, opened 2017-12-07
Validate commands batch size

OndraM:


tests/unit/Model/AssertionTest.php
```diff
@@ -64,4 +65,62 @@ public function provideInvalidTypeIdentifiers(): array
...
+    {
+        $commands = [];
+
+        for ($i = 0; $i < $batchSize; $i++) {
+            $commands[] = ItemPropertySetup::string('name');
+        }
+
+        $this->expectException(DomainException::class);
+        $this->expectExceptionMessage(
+            'Request contains ' . $batchSize . ' commands, but at most 1000 is allowed in one request.'
```
jirinovak, line 110: `sprintf` FTW 😄

jirinovak (dismissed):

tests/unit/Model/AssertionTest.php
OndraM: True, but the simpler tests are (with less logic) the better :). And sprintf is not just smart pattern replacement, it contains logic inside (eg. typecasting).

Quis custodiet ipsos custodes?

foglcz (dismissed):

Issue #17, opened 2017-11-22
Unify self vs static usage


ghost:
Some classes use `new self()` while others use `new static()`. Perhaps it would be a good idea to use `static` everywhere so factory methods work as expected should inheritance get involved.

The same goes for `self::` versus `static::`. Why not use `static::` so static members can be overridden?

Also consider using `@return static` instead of `@return self` so code-completion works in these cases. For example - phpStorm complains about line 14 even though the call would succeed:

![image](https://user-images.githubusercontent.com/33895432/33125534-25dda018-cf81-11e7-9864-ad20b6a99ee9.png)

OndraM:
@Fantomas-7 interested in making this happen? Maybe as part of this sprint?

ghost:
@OndraM Sure. I'll send a PR this week, this should not take long.

PR #56, opened 2017-12-11
Use static instead of self where possible

ghost:
Ref. #17

### DIFF
--- a/src/Matej.php
+++ b/src/Matej.php
@@ -32,6 +32,7 @@ class Matej
         return new RequestBuilderFactory($this->getRequestManager());
     }
 
+    /** @return static */
     public function setHttpClient(HttpClient $client): self
     {
         $this->getRequestManager()->setHttpClient($client);
@@ -39,15 +40,21 @@ class Matej
         return $this;
     }
 
-    /** @internal used mainly in integration tests */
-    public function setBaseUrl($baseUrl): self
+    /**
+     * @internal used mainly in integration tests
+     * @return static
+     */
+    public function setBaseUrl(string $baseUrl): self
     {
         $this->getRequestManager()->setBaseUrl($baseUrl);
 
         return $this;
     }
 
-    /** @codeCoverageIgnore */
+    /**
+     * @codeCoverageIgnore
+     * @return static
+     */
     public function setHttpMessageFactory(MessageFactory $messageFactory): self
     {
         $this->getRequestManager()->setMessageFactory($messageFactory);
@@ -55,7 +62,10 @@ class Matej
         return $this;
     }
 
-    /** @codeCoverageIgnore */
+    /**
+     * @codeCoverageIgnore
+     * @return static
+     */
     public function setHttpResponseDecoder(ResponseDecoderInterface $responseDecoder): self
     {
         $this->getRequestManager()->setResponseDecoder($responseDecoder);
--- a/src/Matej.php
+++ b/src/Matej.php
@@ -32,7 +32,7 @@ class Matej
         return new RequestBuilderFactory($this->getRequestManager());
     }
 
-    /** @return static */
+    /** @return $this */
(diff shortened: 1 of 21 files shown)


foglcz (dismissed):

OndraM (changes requested):
I see few issues there:

1. Where fluent interface is used and `$this` is returned, the best is to declare `$this` as return type as well. Because declaring return type `strict` is in fact not enough - it tells its the same type as class type in current context, but `$this` tells it is also the exact same instance (what it actually is). 
2. This does not solve the original issue mentioned in #17 - extending Commands. For this phpDoc with `@return static` should be added to the named constructors (ie. wherever `return new static(...)` is used).
3. I'm not fan of using constants as "static". Its constant, so we don't expect it to be overwritten, even with LSB, no :)? Why would anyone do this :)? Or is there any usecase on your mind?

ghost:
@OndraM 

1. should be fixed
2. ok, added it there too
3. I don't know.. does it hurt to have this option if the constants are public? But I agree that the use-cases are limited.

src/Matej.php
```diff
@@ -32,30 +32,40 @@ public function request(): RequestBuilderFactory
...
         $this->getRequestManager()->setHttpClient($client);
 
         return $this;
     }
 
-    /** @internal used mainly in integration tests */
-    public function setBaseUrl($baseUrl): self
+    /**
+     * @internal used mainly in integration tests
+     * @return static
```
OndraM: `$this`?

src/Matej.php
```diff
@@ -32,30 +32,40 @@ public function request(): RequestBuilderFactory
...
     {
         $this->getRequestManager()->setBaseUrl($baseUrl);
 
         return $this;
     }
 
-    /** @codeCoverageIgnore */
+    /**
+     * @codeCoverageIgnore
+     * @return static
```
OndraM: dtto, shouldn't it be `$this`?

Also one more case below

src/Model/Command/UserRecommendation.php
```diff
@@ -78,12 +80,14 @@ public function enableHardRotation(): self
     /**
      * Define threshold of how much relevant must the recommended items be to be returned.
      * Default minimal relevance is "low".
+     *
+     * @return static
```
OndraM: Again `$this`? Also below

src/RequestBuilder/AbstractRequestBuilder.php
```diff
@@ -29,6 +29,8 @@
     /**
      * If instance of RequestManager is injected to this builder object, you can build and send the request directly
      * via send() method of the builder itself.
+     *
+     * @return static
```
OndraM: $this here and below :)

OndraM (changes requested):
I believe the `$this` should be used in typehints wherever fluent interface is used and `$this` is actually returned. It will still work with `$this` (instead of `static`) to cover the use case mentioned in #17 (code completion on extended classes), right?

src/Model/Command/UserRecommendation.php
ghost: `@return static` and `@return $this` mean the same thing 🤔

OndraM: `static` tells its just the same *type* (resolved by LSB).
`$this` tells its not only the same type, but also the same actual instance. This means $this is "subset" of static, but not the same thing, and cannot be interchanged.

Its also recognized like this in proposed phpdoc PSR standard, see https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#keyword - points 14. and 15.

ghost: Cool, didn't know that. Fixed.

OndraM (approved):
👍 Thanks!

Issue #8, opened 2017-11-13
Make Response class iterable for CommandResponses


OndraM:
Consider implementing eg. SplObjectStorage into `Lmc\Matej\Model\Response` to allow simplier acess to commandResponses.

However set/unset methods must be not implement (ie. throw an exception), as the object is not allowed to be modified after creation.

OndraM:
No longer relevant/needed.

PR #57, opened 2017-12-11
Add syntax sugar for Sorting and Recommendation endpoints (RAD-746)

foglcz:
So now you can do this:

```php
$recommendations = $matej->request()
    ->recommendation(UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600))
    ->send()
    ->getRecommendation()
    ->getData();

$sortedItems =  $matej->request()
    ->sorting(Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3']))
    ->send()
    ->getSorting()
    ->getData();

$properties = $matej->request()
    ->getItemProperties()
    ->send()
    ->getData();
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ### Changed
 - Validate max. 1000 commands are added to `campaign()`, `events()`, `setupItemProperties()` and `deleteItemProperties()` requests (in accordance with Matej batch API limit).
+- `Response` now contains `getCommandResponse(int $index)` to access individual command responses.
+- `sorting()->send()` request now returns `SortingResponse` instance, which provides semantic access to data of the response (ie `->getSorting()`).
+- `recommendation()->send()` request now returns `RecommendationResponse` instance, which provides semantic access to data of the response (ie. `->getRecommendation()`).
+- `getItemProperties()->send()` request now returns `ItemPropertiesListResponse` instance, which provides semantic access to data of the response (ie. `->getData()`).
 
 ## 1.0.0 - 2017-12-05
 ### Added
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ foreach ($response->getCommandResponses() as $commandResponse) {
 }
 ```
 
+[Recommendation](#recommendations-for-single-user), [Sorting](#request-item-sorting-for-single-user)
+and [Item Properties](#item-properties-setup-to-setup-you-matej-database) endpoints have syntax sugar,
+which makes processing responses easier. See below for detailed examples.
+
 ### Item properties setup (to setup you Matej database)
 
 ```php
@@ -95,6 +99,8 @@ $response = $matej->request()
 $response = $matej->request()
     ->getItemProperties()
     ->send();
+
+$properties = $response->getData();
 
 // Delete item property from database:
 $response = $matej->request()
@@ -144,6 +150,8 @@ $response = $matej->request()
     ->setInteraction(Interaction::purchase('user-id', 'item-id')) // optional
     ->setUserMerge(UserMerge::mergeInto('user-id', 'source-id')) // optional
     ->send();
+
+$recommendations = $response->getRecommendation()->getData();
 ```
 
 You can also set more granular options of the recommendation command:
@@ -157,6 +165,20 @@ $recommendation->setFilters(['valid_to >= NOW']) // Note this filter is present 
 $response = $matej->request()
     ->recommendation($recommendation)
     ->send();
+```
+
+From `$response`, you can also access rest of the data:
+
+```php
+$response = $matej->request()
+    ->recommendation($recommendation)
+    ->send();
+
+echo $response->getInteraction()->getStatus();    // SKIPPED
+echo $response->getUserMerge()->getStatus();      // SKIPPED
+echo $response->getRecommendation()->getStatus(); // OK
+
+$recommendations = $response->getRecommendation()->getData();
 ```
 
 ### Request item sorting for single user
@@ -173,6 +195,21 @@ $response =  $matej->request()
     ->setUserMerge(UserMerge::mergeInto('user-id', 'source-id')) // optional
     ->send();
 
+$sortedItems = $response->getSorting()->getData();
+```
+
+From `$response`, you can also access rest of the data:
+
+```php
+$response = $matej->request()
+    ->sorting($sorting)
+    ->send();
+
+echo $response->getInteraction()->getStatus(); // SKIPPED
+echo $response->getUserMerge()->getStatus();   // SKIPPED
+echo $response->getSorting()->getStatus();     // OK
+
+$sortedData = $response->getSorting()->getData();
 ```
 
 ### Request batch of recommendations/item sortings
--- a/src/Model/Assertion.php
+++ b/src/Model/Assertion.php
@@ -3,6 +3,7 @@
 namespace Lmc\Matej\Model;
 
 use Lmc\Matej\Exception\DomainException;
+use Lmc\Matej\Exception\LogicException;
 use Lmc\Matej\Model\Command\AbstractCommand;
 
 /**
@@ -37,13 +38,28 @@ class Assertion extends \Assert\Assertion
      *
      * @param AbstractCommand[] $commands
      */
-    public static function batchSize(array $commands)
+    public static function batchSize(array $commands): bool
     {
         static::lessOrEqualThan(
             count($commands),
             self::MAX_BATCH_SIZE,
             'Request contains %s commands, but at most %s is allowed in one request.'
         );
+
+        return true;
+    }
+
+    /**
+     * Assert that provided classname is an instance or subclass of Response
+     */
+    public static function isResponseClass(string $wantedClass): bool
+    {
+        if (
+            !is_a($wantedClass, Response::class, true) &&
+            !is_subclass_of($wantedClass, Response::class)
+        ) {
+            throw LogicException::forClassNotExtendingOtherClass($wantedClass, Response::class);
+        }
 
         return true;
     }
--- a/src/Model/Request.php
+++ b/src/Model/Request.php
@@ -17,13 +17,16 @@ class Request
     private $data;
     /** @var string */
     private $requestId;
+    /** @var string */
+    private $responseClass;
 
-    public function __construct(string $path, string $method, array $data = [], string $requestId = null)
+    public function __construct(string $path, string $method, array $data = [], string $requestId = null, string $responseClass = Response::class)
     {
         $this->path = $path;
         $this->method = $method;
         $this->data = $data;
         $this->requestId = $requestId ?? Uuid::uuid4()->toString();
+        $this->setResponseClass($responseClass);
     }
 
     public function getPath(): string
@@ -44,5 +47,19 @@ class Request
     public function getRequestId(): string
     {
         return $this->requestId;
+    }
+
+    public function getResponseClass(): string
+    {
+        return $this->responseClass;
+    }
+
+    private function setResponseClass(string $class): self
+    {
+        Assertion::isResponseClass($class);
+
+        $this->responseClass = $class;
+
+        return $this;
     }
 }
--- a/src/Model/Response.php
+++ b/src/Model/Response.php
@@ -79,6 +79,14 @@ class Response
     }
 
     /**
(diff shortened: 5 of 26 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 ### Changed
 - Validate max. 1000 commands are added to `campaign()`, `events()`, `setupItemProperties()` and `deleteItemProperties()` requests (in accordance with Matej batch API limit).
+- `Lmc\Matej\Model\Response` now contains `getCommandResponse(int $index)` to access individual responses
+- `sorting()->send()` request now returns `Matej\Model\Response\SortingResponse`
```
jirinovak: `instance` for consistency or maybe it's not necessary at all

README.md
```diff
@@ -159,6 +165,21 @@ $response = $matej->request()
...
+$response = $matej->request()
+    ->recommendation($recommendation)
+    ->send();
+
+echo $response->getInteraction()->getStatus();    // SKIPPED
+echo $response->getUserMerge()->getStatus();      // SKIPPED
+echo $response->getRecommendation()->getStatus(); // OK
+
+$recommendations = $response->getRecommendation()->getData();
+dump($recommendations);
```
jirinovak: What this dump is good for?

Same applies to sorting section.

src/Exception/InvalidArgumentException.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Lmc\Matej\Exception;
+
+/**
+ * Exception represents error in the program logic.
+ */
+class InvalidArgumentException extends \InvalidArgumentException implements MatejExceptionInterface
+{
+    public static function forClassNotExtendingOtherClass($class, $wantedClass)
+    {
+        return new self(sprintf('Class %s has to be either instance of, or subclass of %s', $class, $wantedClass));
```
jirinovak: Maybe provided class would be also useful information.

`subclass of %s. But %s class provided` or similarly.

src/Model/Request.php
```diff
@@ -17,13 +17,16 @@ class Request
     private $data;
     /** @var string */
     private $requestId;
+    /** @var string */
+    private $responseClass;
 
-    public function __construct(string $path, string $method, array $data = [], string $requestId = null)
+    public function __construct(string $path, string $method, array $data = [], string $requestId = null, string $responseClass = null)
```
jirinovak: I don't like idea that request knows about response. This should be responsibility of `ResponseDecoder`.

README.md
```diff
@@ -59,9 +59,9 @@ $response = $matej->request()
     ...
 ```
 
-See below for examples of building request for each endpoint.
-
-To process the response:
+In general, all requests to Matej execute in batch format. This means that a request for single `/recommendation` is in fact
+a single request, which contains up to three `Matej\Model\Command\AbstractCommands`, that are indexed as per Matejs documentation.
```
OndraM: This this part is a bit confusing. Without deeper knowledge of the API, you may not have a clue why you talk about three commands. 

Maybe this part could be kept as-is, only examples in the endpoints documentation (plus the note below) could be enough...

README.md
```diff
@@ -79,6 +79,9 @@ foreach ($response->getCommandResponses() as $commandResponse) {
 }
 ```
 
+Some endpoints have facades, that allow you to work more easily with that particular endpoint.
```
OndraM: The involved endpoints could be named there (with hyperlink to the header).

Also I'd not call it "facade"... Its imho more "shortcut" or "syntax sugar".

src/Model/Request.php
OndraM: Well, the request knows what structure of response it is supposed to receive, no?

src/Model/Response.php
```diff
@@ -78,6 +78,14 @@ public function getNumberOfSkippedCommands(): int
         return $this->numberOfSkippedCommands;
     }
 
+    /**
+     * Return individual command response by its index (0 indexed)
+     */
+    public function getCommandResponse(int $index): CommandResponse
```
OndraM, line 84: Wouldn't be better implementing https://github.com/lmc-eu/matej-client-php/issues/8 ?

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -46,6 +46,7 @@ public function shouldBuildRequestWithCommands(): void
         $this->assertSame($recommendationsCommand, $requestData[2]);
 
         $this->assertSame('custom-request-id-foo', $request->getRequestId());
+        $this->assertSame(Response\RecommendationsResponse::class, $request->getResponseClass());
```
OndraM: Please import the whole namespace :). Also below.

src/Model/Assertion.php
```diff
@@ -47,4 +48,19 @@ public static function batchSize(array $commands)
 
         return true;
     }
+
+    /**
+     * Assert that provided classname is an instance or subclass of Response
+     */
+    public static function isResponseClass(string $wantedClass): bool
```
OndraM, line 55: Is there a unit test for this?

OndraM (changes requested):

src/Model/Assertion.php
foglcz: it was missing, added. (edit: the test itself was carried out with Request/Response, but there was no test for assertion directly.)

src/Model/Response.php
foglcz: Depends. I think a direct accessor function is desirable.

As for the object storage; I'm inclined to implement `Countable`, `Iterator` and `ArrayAccess` directly, so that you can:

```php
$response[2]->getData();
```

In that case, I'd still prefer a function to get the data as well, as it feels less magical :)

src/Model/Request.php
foglcz: +1. Also ResponseDecoder gets only the `PSR ResponseInterface` as parameter, so we'd have to put the pairing between requests & responses into `RequestManager`.

To me it feels better to pair a response directly with request.

src/Model/Request.php
```diff
@@ -17,13 +17,16 @@ class Request
...
     {
         $this->path = $path;
         $this->method = $method;
         $this->data = $data;
         $this->requestId = $requestId ?? Uuid::uuid4()->toString();
+        $this->setResponseClass($responseClass ?? Response::class);
```
OndraM: Why don't use `Response::class` as default value, instead of doing this `??`? Unlike the Uuid runtime stuff above this is IMHO not needed. The default value would be also clearly visible in method signature in IDE.

tests/unit/Model/AssertionTest.php
```diff
@@ -123,4 +124,17 @@ public function provideInvalidBatchSize(): array
             [2000],
         ];
     }
+
+    /** @test */
+    public function shouldAssertResponseClass(): void
```
OndraM: And do you also test the "happy path"? Ie. when you actually use the Response class/subclass name? Maybe create some dummy fixture class extending the Response?

tests/unit/RequestBuilder/SortingRequestBuilderTest.php
```diff
@@ -46,6 +46,7 @@ public function shouldBuildRequestWithCommands(): void
         $this->assertSame($sortingCommand, $requestData[2]);
 
         $this->assertSame('custom-request-id-foo', $request->getRequestId());
+        $this->assertSame(Response\SortingResponse::class, $request->getResponseClass());
```
OndraM: Please import the whole namespace and use only the short class name :)

README.md
```diff
@@ -92,9 +96,10 @@ $response = $matej->request()
     ->send();
 
 // Get list of item properties that are defined in matej
-$response = $matej->request()
+$properties = $matej->request()
     ->getItemProperties()
-    ->send();
+    ->send()
```
OndraM: I know its shorter, but I'd prefer splitting the examples into two variables, and don't mix the fluent interface like this (where the return type changes during the calls) , whats is IMO not the best practice and not really transparent.

```
$response = $matej->request()
     ->getItemProperties()
     ->send();

$properties = $response->getData();
```

Similarly for the other examples...
```
$response = $matej->request()
     ->recommendation(UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600))
     ->setInteraction(Interaction::purchase('user-id', 'item-id')) // optional
     ->setUserMerge(UserMerge::mergeInto('user-id', 'source-id')) // optional
     ->send();

$recommendedItems  = $response->getRecommendation()
    ->getData();
```

etc.?

README.md
```diff
@@ -167,12 +188,28 @@ and user merge event in one request, to make them taken into account when execut
...
 $matej = new Matej('accountId', 'apikey');
 
-$response =  $matej->request()
+$sortedData =  $matej->request()
     ->sorting(Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3']))
     ->setInteraction(Interaction::purchase('user-id', 'item-id')) // optional
     ->setUserMerge(UserMerge::mergeInto('user-id', 'source-id')) // optional
+    ->send()
+    ->getSorting()
+    ->getData();
```
OndraM: And this could be 

```
$response =  $matej->request()
     ->sorting(Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3']))
     ->setInteraction(Interaction::purchase('user-id', 'item-id')) // optional
     ->setUserMerge(UserMerge::mergeInto('user-id', 'source-id')) // optional
     ->send();

$sortedItems  = $response->getSorting()
    ->getData();
```

Your thoughts :)?

src/Exception/InvalidArgumentException.php
```diff
@@ -0,0 +1,14 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Exception;
+
+/**
+ * Exception represents error in the program logic.
+ */
+class InvalidArgumentException extends \InvalidArgumentException implements MatejExceptionInterface
```
OndraM: Do we need to introduce another exception type? This is IMO edge case, so it may be unnecessary to have a specific exception for this - just extending LogicException would be OK for me, and its more general, but semantically still fine (InvalidArgumentException is in SPL subtype of LogicException).

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 ### Changed
 - Validate max. 1000 commands are added to `campaign()`, `events()`, `setupItemProperties()` and `deleteItemProperties()` requests (in accordance with Matej batch API limit).
+- `Lmc\Matej\Model\Response` now contains `getCommandResponse(int $index)` to access individual responses
```
OndraM: Please could you add dots at each EOL (for consistency)?

OndraM: Also maybe no need to write FQCN... "`Response` now contains" ... is IMHO sufficiently unambiguous. Similarly for following lines.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
...
+- `sorting()->send()` request now returns `Matej\Model\Response\SortingResponse` instance
```
OndraM: "`sorting()->send()` request now returns `SortingResponse` instance which provides shortcut methods." or something like this?

README.md
```diff
@@ -159,6 +166,20 @@ $response = $matej->request()
     ->send();
 ```
 
+`send()` method returns a Recommendations response, which will allows you to access full response data:
```
OndraM: Not sure here, but shouldn't it be "which will allow you"?

OndraM: And maybe "which will allow you to directly access response data:" (to emphasize the difference from using `getCommandResponse()` / `getCommandResponses()`)

(Same for the Sorting request below)

OndraM (changes requested):

OndraM (approved):

src/Model/Response/ItemPropertiesListResponse.php
```diff
@@ -0,0 +1,28 @@
+<?php declare(strict_types=1);
```
OndraM, line 1: Oh, i just noted we are missing UT for these responses... Could you please add them?

OndraM (changes requested):
Also the code-coverage reporting was broken, I fixed it in master, so could you rebase the branch or merge master branch here to get the coveralls reporting working again?

OndraM (approved):

PR #58, opened 2017-12-12
Fix default filter in user recommendations

foglcz:
At this moment, filters are hardcoded. There is no `valid_to >= now` filter, we have `NOW <= valid_to`

This fixes the issue - the filter should work now.

Filter query language is being implemented under RAD-286, and should be available soon.


ghost (approved):

foglcz:
Closing, not needed.

PR #60, opened 2017-12-20
Fix checking of user ids when building requests for recommendation and sorting

foglcz:
Sorting and Recommendation endpoints allow optional Interaction and User Merge. These are currently checked, so that they contain the recommendation/sorting user id.

When user merge is present, the `interaction` has to cointain **old user id**, not the new one.

```
Consider /recommendation or /sorting request:
([interaction], [user merge], recommendation|sorting)

These requests are valid:
(A, null, A)
(A, A -> B, B)
(null, A -> B, B)

These requests are invalid and should be rejected:
( A  ,  null , B)
( A  , B -> A, A)
( B  , B -> A, B)
``` 

Currently, Matej's API will accept valid requests, and reject the wrong ones.

Current PHP Client behaviour:
```
(A, A -> B, B) - throws exception (interaction differs from master user)
(A, B -> A, A) - passes validation, but Matej's API will reject this
```

This PR remedies the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 ### Changed
 - Unify `self`/`static`/`$this` return typehints.
 
+### Fixed
+- Checking of consistent user ids on `recommendation()` and `sorting()` requests was not working in accordance with API behavior.
+
 ## 1.1.0 - 2017-12-20
 ### Added
 - Endpoint to get all defined item properties in the Matej database (`$matej->request()->getItemProperties()`).
--- a/src/Exception/LogicException.php
+++ b/src/Exception/LogicException.php
@@ -23,6 +23,17 @@ class LogicException extends \LogicException implements MatejExceptionInterface
         return new static($message);
     }
 
+    public static function forInconsistentUserMergeAndInteractionCommand($userMergeId, $interactionUserId)
+    {
+        $message = sprintf(
+            'Source user in UserMerge command ("%s") must be the same as user in Interaction command ("%s")',
+            $userMergeId,
+            $interactionUserId
+        );
+
+        return new self($message);
+    }
+
     public static function forClassNotExtendingOtherClass($class, $wantedClass)
     {
         return new self(sprintf('Class %s has to be instance or subclass of %s.', $class, $wantedClass));
--- a/src/RequestBuilder/RecommendationRequestBuilder.php
+++ b/src/RequestBuilder/RecommendationRequestBuilder.php
@@ -47,7 +47,8 @@ class RecommendationRequestBuilder extends AbstractRequestBuilder
 
     public function build(): Request
     {
-        $this->assertConsistentUsersInCommands();
+        $this->assertInteractionUserId();
+        $this->assertUserMergeUserId();
 
         return new Request(
             static::ENDPOINT_PATH,
@@ -58,14 +59,41 @@ class RecommendationRequestBuilder extends AbstractRequestBuilder
         );
     }
 
-    private function assertConsistentUsersInCommands(): void
+    /**
+     * Assert that interaction user ids are ok:
+     * - (A,  null,  A)
+     * - (A, A -> ?, ?)
+     */
+    private function assertInteractionUserId(): void
     {
-        $mainCommandUser = $this->userRecommendationCommand->getUserId();
-        if ($this->interactionCommand !== null && $mainCommandUser !== $this->interactionCommand->getUserId()) {
+        if ($this->interactionCommand === null) {
+            return;
+        }
+
+        $interactionUserId = $this->interactionCommand->getUserId();
+
+        // (A, null, A)
+        if ($this->userMergeCommand === null && $interactionUserId !== $this->userRecommendationCommand->getUserId()) {
             throw LogicException::forInconsistentUserId($this->userRecommendationCommand, $this->interactionCommand);
         }
 
-        if ($this->userMergeCommand !== null && $mainCommandUser !== $this->userMergeCommand->getUserId()) {
+        // (A, A -> ?, ?)
+        if ($this->userMergeCommand !== null && $interactionUserId !== $this->userMergeCommand->getSourceUserId()) {
+            throw LogicException::forInconsistentUserMergeAndInteractionCommand(
+                $this->userMergeCommand->getSourceUserId(),
+                $interactionUserId
+            );
+        }
+    }
+
+    /**
+     * Assert user merge id is ok:
+     * (?, ? -> A, A)
+     */
+    private function assertUserMergeUserId(): void
+    {
+        if ($this->userMergeCommand !== null
+            && $this->userMergeCommand->getUserId() !== $this->userRecommendationCommand->getUserId()) {
             throw LogicException::forInconsistentUserId($this->userRecommendationCommand, $this->userMergeCommand);
         }
     }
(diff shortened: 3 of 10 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 <!-- There is always Unreleased section on the top. Subsections (Added, Changed, Fixed, Removed) should be added as needed. -->
 
 ## Unreleased
+- Fixed checking of user ids on `recommendation()` and `sorting()` requests
```
OndraM: Dot on EOL :)

src/RequestBuilder/RecommendationRequestBuilder.php
```diff
@@ -58,13 +58,24 @@ public function build(): Request
...
-        $mainCommandUser = $this->userRecommendationCommand->getUserId();
-        if ($this->interactionCommand !== null && $mainCommandUser !== $this->interactionCommand->getUserId()) {
-            throw LogicException::forInconsistentUserId($this->userRecommendationCommand, $this->interactionCommand);
+        $interactionUser = $this->interactionCommand ? $this->interactionCommand->getUserId() : null;
+        $mergeFromUser = $this->userMergeCommand ? $this->userMergeCommand->getSourceUserId() : null;
+        $mergeIntoUser = $this->userMergeCommand ? $this->userMergeCommand->getUserId() : null;
+        $recommendationUser = $this->userRecommendationCommand->getUserId();
+
+        // ([A], [A -> ?], ?)
+        if ($this->interactionCommand !== null && $this->userMergeCommand !== null && $interactionUser !== $mergeFromUser) {
```
OndraM: I would separate those three conditions inside `if` to a private method (for better readability).

Similary in the cases below (and in other Builders).

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -108,4 +108,39 @@ public function shouldThrowExceptionWhenUserOfUserMergeDiffersFromSorting(): voi
...
+         * Should fail: ([interaction], [user merge], [recommendation]):
+         *              (A, B -> A, A)
+         */
+        $interactionCommand = Interaction::purchase('old-user-id', 'test-item-id');
+        $userMergeCommand = UserMerge::mergeInto('old-user-id', 'new-user-id');
+        $recommendationsCommand = UserRecommendation::create('old-user-id', 5, 'scenario', 0.5, 3600);
+
+        $this->expectException(LogicException::class);
+        $this->expectExceptionMessage(
+            'User in interaction ("old-user-id") must be the same as user in UserMerge::from ("new-user-id")'
```
OndraM: Isn't the `UserMerge::from` confusing? Would something like `User in interaction ("old-user-id") must be the same as the source user of UserMerge ("new-user-id")` better?

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -108,4 +108,39 @@ public function shouldThrowExceptionWhenUserOfUserMergeDiffersFromSorting(): voi
...
+        $interactionCommand = Interaction::purchase('old-user-id', 'test-item-id');
+        $userMergeCommand = UserMerge::mergeFromSourceToTargetUser('old-user-id', 'new-user-id');
+        $recommendationsCommand = UserRecommendation::create('new-user-id', 5, 'scenario', 0.5, 3600);
+
+        $builder = new RecommendationRequestBuilder($recommendationsCommand);
+        $builder->setUserMerge($userMergeCommand);
+        $builder->setInteraction($interactionCommand);
+        $this->assertInstanceOf(Request::class, $builder->build());
+
+        /**
```
OndraM: Separate this into standalone test method? (And the test could be renamed then)

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -108,4 +108,39 @@ public function shouldThrowExceptionWhenUserOfUserMergeDiffersFromSorting(): voi
...
+         * Should fail: ([interaction], [user merge], [recommendation]):
+         *              (A, B -> A, A)
+         */
+        $interactionCommand = Interaction::purchase('old-user-id', 'test-item-id');
+        $userMergeCommand = UserMerge::mergeInto('old-user-id', 'new-user-id');
```
OndraM: The data naming is confusing - `mergeInto` params are `$targetUserId, $sourceUserIdToBeDeleted`. So call them `old-user-id` and `new-user-id` (while the later one is in fact the one to be deleted, ie. not the "new" user id) may lead to confusion imo.

tests/unit/RequestBuilder/SortingRequestBuilderTest.php
```diff
@@ -101,4 +101,39 @@ public function shouldThrowExceptionWhenUserOfUserMergeDiffersFromSorting(): voi
...
+        $interactionCommand = Interaction::purchase('old-user-id', 'test-item-id');
+        $userMergeCommand = UserMerge::mergeFromSourceToTargetUser('old-user-id', 'new-user-id');
+        $sortingCommand = Sorting::create('new-user-id', ['itemId1', 'itemId2']);
+
+        $builder = new SortingRequestBuilder($sortingCommand);
+        $builder->setUserMerge($userMergeCommand);
+        $builder->setInteraction($interactionCommand);
+        $this->assertInstanceOf(Request::class, $builder->build());
+
+        /**
```
OndraM: Dtto, put this into standalone test?

tests/unit/RequestBuilder/SortingRequestBuilderTest.php
```diff
@@ -101,4 +101,39 @@ public function shouldThrowExceptionWhenUserOfUserMergeDiffersFromSorting(): voi
...
+         * Should fail: ([interaction], [user merge], [sorting]):
+         *              (A, B -> A, A)
+         */
+        $interactionCommand = Interaction::purchase('old-user-id', 'test-item-id');
+        $userMergeCommand = UserMerge::mergeInto('old-user-id', 'new-user-id');
+        $sortingCommand = Sorting::create('new-user-id', ['itemId1', 'itemId2']);
+
+        $this->expectException(LogicException::class);
+        $this->expectExceptionMessage(
+            'User in interaction ("old-user-id") must be the same as user in UserMerge::from ("new-user-id")'
```
OndraM: Dtto, i suggest rewording the `UserMerge::from` part :)

src/Exception/LogicException.php
```diff
@@ -10,13 +10,23 @@
...
+    {
+        return self::forInconsistentRawUserId(
+            $mainCommand->getUserId(),
+            (new \ReflectionClass($mainCommand))->getShortName() . ' command',
+                $additionalCommand->getUserId(),
+            (new \ReflectionClass($additionalCommand))->getShortName() . ' command'
+        );
+    }
+
+    public static function forInconsistentRawUserId($mainUserId, $mainSubject, $additionalUserId, $additionalSubject)
```
OndraM: Maybe in the two cases where this method is used directly, it could be replaced with own specific method with less generic name (as the mainSubject and additionalSubject values are always the same when it is called directly).

OndraM (changes requested):

OndraM (dismissed):

PR #64, opened 2018-02-21
Add endpoint for database reset

OndraM:
Following today's release of RAD-898, we can finally implement interface for simple database-reset.

Usage is shown in README.

I'm thinking about the name of PlainResponse (?). Which is supposed to be generic type for responses having always exactly one response command (currently database reset and get item properties), like this:

```js
{
  "commands": {
    "number_of_commands": 1,
    "number_of_failed_commands": 0,
    "number_of_skipped_commands": 0,
    "number_of_successful_commands": 1,
    "responses": [
      {
        "data": [
            // there may or may be not something
        ],
        "message": "",
        "status": "OK"
      }
    ]
  },
  "message": "",
  "status": "OK"
}
```

Is `PlainResponse` the best name? Any other suggestions? Like `SingleCommandResponse`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 <!-- There is always Unreleased section on the top. Subsections (Added, Changed, Fixed, Removed) should be added as needed. -->
 
 ## Unreleased
+### Added
+- Endpoint to reset Matej test-databases (`$matej->request()->resetDatabase()`).
+
 ### Fixed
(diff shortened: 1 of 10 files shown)


jirinovak:
> Is PlainResponse the best name? Any other suggestions? Like SingleCommandResponse?

Other response class names says also nothing about command count. So I would stick with `PlainResponse`.

jirinovak (approved):

PR #67, opened 2018-04-11
Implement AB testing for Matej recommendation models

foglcz:
|  |  |
| --- | --- |
| **Type** | Feature |
| **Fixes issues** | RAD-1066 |
| **BC Break** | no |
| **Tests updated** | yes |

Contains implementation for A/B testing of various recommendation models. These models have to be implemented/deployed in Matej first, before they're available for any given client. This is by design, as new model requires full recalculation, and therefore availability is individual depending on the size of clients data.

By default, only `default` model is available for every client (which means *clients* default, not *matej* default.)

Requires **Matej release ~v7.37~ v7.36**
Currently deployed to Staging instance, which contains models `default` and `model_b`.

There is no API endpoint that'd list available models. No validation beyond basic identifier is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,51 @@ $response = $matej->request()
     ->send();
 ```
 
+### A/B Testing support
+`Recommendation` and `Sorting` commands now support optional A/B testing of various models. This has to be set up in Matej first,
+but once available, you can specify which model you want to use when requesting recommendations:
+
+```php
+$response = $matej->request()
+    ->recommendation(UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600, 'custom_name_of_model_b'))
+    ->send();
+```
+
+Or, you can use `setModelName` setter, if you're specifying parameters granularly:
+
+```php
+$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
+$recommendation->setModelName('custom_name_of_model_b');
+
+$response = $matej->request()
+    ->recommendation($recommendation)
+    ->send();
+```
+
+Similarly, you can do that for `Sorting` request:
+
+```php
+$response =  $matej->request()
+    ->sorting(Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3'], 'custom_name_of_model_b'))
+    ->send()
+;
+
+// ... or ...
+$sorting = Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3']);
+$sorting->setModelName('custom_name_of_model_b');
+
+$response = $matej->request()
+    ->sorting($sorting)
+    ->send();
+```
+
+`custom_name_of_model_b` will be provided to you by LMC.
+
+If you use `NULL`, or `'default'`, the default model for your instance will be used.
+
+A/B testing is available in `/recommendation` and `/sorting` endpoints for individual requests,
+as well as in `/campaign` endpoint for batch requests.
+
 ### Exceptions and error handling
 
 Exceptions are thrown only if the whole Request to Matej failed (when sending, decoding, authenticating etc.) or if
--- a/src/Model/Command/Sorting.php
+++ b/src/Model/Command/Sorting.php
@@ -14,11 +14,17 @@ class Sorting extends AbstractCommand implements UserAwareInterface
     private $userId;
     /** @var string[] */
     private $itemIds = [];
+    /** @var string|null */
+    private $modelName = null;
 
-    private function __construct(string $userId, array $itemIds)
+    private function __construct(string $userId, array $itemIds, string $modelName = null)
     {
         $this->setUserId($userId);
         $this->setItemIds($itemIds);
+
+        if ($modelName !== null) {
+            $this->setModelName($modelName);
+        }
     }
 
     /**
@@ -26,9 +32,23 @@ class Sorting extends AbstractCommand implements UserAwareInterface
      *
      * @return static
      */
-    public static function create(string $userId, array $itemIds): self
+    public static function create(string $userId, array $itemIds, string $modelName = null): self
     {
-        return new static($userId, $itemIds);
+        return new static($userId, $itemIds, $modelName);
+    }
+
+    /**
+     * Set A/B model name
+     *
+     * @return $this
+     */
+    public function setModelName(string $modelName): self
+    {
+        Assertion::typeIdentifier($modelName);
+
+        $this->modelName = $modelName;
+
+        return $this;
     }
 
     public function getUserId(): string
@@ -57,9 +77,15 @@ class Sorting extends AbstractCommand implements UserAwareInterface
 
     protected function getCommandParameters(): array
     {
-        return [
+        $parameters = [
             'user_id' => $this->userId,
             'item_ids' => $this->itemIds,
         ];
+
+        if ($this->modelName !== null) {
+            $parameters['model_name'] = $this->modelName;
+        }
+
+        return $parameters;
     }
 }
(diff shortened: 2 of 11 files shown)


src/Model/Command/Sorting.php
```diff
@@ -14,21 +14,40 @@ class Sorting extends AbstractCommand implements UserAwareInterface
     private $userId;
     /** @var string[] */
     private $itemIds = [];
+    /** @var string|null */
+    private $modelName = null;
 
-    private function __construct(string $userId, array $itemIds)
+    private function __construct(string $userId, array $itemIds, ?string $modelName = null)
```
jirinovak, line 20: Maybe we can use `default` as default value and request `string`. As a benefit we can get rid of all null checks in other methods.

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
```diff
@@ -6,13 +6,37 @@
...
+use Lmc\Matej\Model\Command\ItemPropertySetup;
 use Lmc\Matej\Model\Command\UserMerge;
+use Lmc\Matej\RequestBuilder\ItemPropertiesSetupRequestBuilder;
 
 /**
  * @covers \Lmc\Matej\RequestBuilder\EventsRequestBuilder
  */
 class EventsRequestBuilderTest extends IntegrationTestCase
 {
+    private function buildAndSendPropertySetupRequest(ItemPropertiesSetupRequestBuilder $builder): void
```
jirinovak, line 18: Commit 95c55be86ff6a93469e0251226ea8393f2486bd0 seems to be unrelated with this PR and should be separate PR. Or am I wrong?

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
```diff
@@ -6,13 +6,37 @@
...
+    {
+        $builder->addProperty(ItemPropertySetup::string('test_property_a'))
+            ->addProperty(ItemPropertySetup::string('test_property_b'))
+            ->addProperty(ItemPropertySetup::string('test_property_c'))
+            ->send();
+    }
+
+    protected function setup(): void
```
jirinovak, line 26: `setup` and `tearDown` methods should be first.

And helper method `buildAndSendPropertySetupRequest` should be at the end of test class.

jirinovak (changes requested):

kubasimon:
What is the motivation to make client responsible for selecting right 'model' (a/b testing) and not to select right model (or a/b test models) on server? (especially when new model "by design requires full recalculation")

VincTheSecond:
@tenerd Currently, Matej is not able to measure business metrics (click rates, conversions) exactly as it requires clients to send us something like RPS codes for all interactions. In 2018.Q2 we will introduce a new strategy how to do it. After that, we will able to provide AB testing exclusively in Matej, without any client participation. For now, this is a temporary solution for clients who want to test different models even in 2018.Q2 :-)

README.md
```diff
@@ -251,6 +251,51 @@ $response = $matej->request()
...
+$sorting->setModelName('custom_name_of_model_b');
+
+$response = $matej->request()
+    ->sorting($sorting)
+    ->send();
+```
+
+`custom_name_of_model_b` will be provided to you by LMC.
+
+If you use `NULL`, or `'default'`, the default model for your instance will be used.
```
OndraM, line 294: `null` in lowercase

README.md
```diff
@@ -251,6 +251,51 @@ $response = $matej->request()
...
+
+A/B testing is available in `/recommendation` and `/sorting` endpoints for individual requests,
```
OndraM, line 296: I'd not refer to endpoints in the "URL" form, rather this could by hyperlink to the respective query builders:
`... is available in [Recommendation](#recommendations-for-single-user), ..."`

README.md
```diff
@@ -251,6 +251,51 @@ $response = $matej->request()
...
+as well as in `/campaign` endpoint for batch requests.
```
OndraM, line 297: "... as well as in bath request created via `campaign()` builder."

src/Model/Command/Sorting.php
OndraM: The question is whether Matej behavior will be the same for `null` and `"default"`, @foglcz ?

src/Model/Command/Sorting.php
```diff
@@ -14,21 +14,40 @@ class Sorting extends AbstractCommand implements UserAwareInterface
...
+        $this->setModelName($modelName);
     }
 
     /**
      * Sort given item ids for user-based recommendations.
      *
      * @return static
      */
-    public static function create(string $userId, array $itemIds): self
+    public static function create(string $userId, array $itemIds, string $modelName = null): self
```
OndraM, line 32: What about adding second static builder (with modelName) and keeping this one as is? `createWithModel()` etc.

Because Vincent wrote this is temporary solution, to be removed in the future once implemented inside Matej. And it is easier to remove (and find in codebase) on particular method call than method call with third parameter.

src/Model/Command/UserRecommendation.php
```diff
@@ -51,16 +54,18 @@ private function __construct(string $userId, int $count, string $scenario, float
      * Set from 0.0 for no rotation (same items will be recommended) up to 1.0 (same items should not be recommended).
      * @param int $rotationTime Specify for how long will the item's rotationRate be taken in account and so the item
      * is penalized for recommendations.
+     * @param string|null $modelName Specify which model you want to use
```
OndraM, line 57: Dtto, what do you think about introducing second constructor?

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
OndraM: Also, why is the test failing 🤔?

tests/integration/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -20,7 +21,7 @@ public function shouldExecuteRecommendationRequestOnly(): void
     {
         $response = $this->createMatejInstance()
             ->request()
-            ->recommendation($this->createRecommendationCommand('user-a'))
+            ->recommendation($this->createRecommendationCommand('user-a', null))
```
OndraM, line 24: I think this should be kept without the second param - to test the "basic" use-case (ie. called wihout optional params).

OndraM (changes requested):

src/Model/Command/Sorting.php
foglcz: On most clients, it will be - until they decide to use different model than default. `model_name` is optional parameter in the payload, allowing us to switch default models server-side.

For example, if the client decides to test default model against new model (`alpha`), and then decides to use the `alpha` as new default, then we won't *rename* the model -- we'd just switch configuration server side to use the new one for all requests that don't explicitly request a different model name.

Changing parameters of existing model needs full recalculation of the model - which is a costly operation. By the time the client decides to change the default model, we'll already have everything calculated & up to date in the new model; so doing a recalc feels kinda pointless if we can just change the default pointer server-side.

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
foglcz: ahaha, that's because fresh instances don't have these item properties set up.
I can cherry-pick it into separate PR if you want. I don't really feel it's a problem though.

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
foglcz: :+1: will fix, thanks

foglcz:
Closing based on review - this PR will be split into two.

src/Model/Command/UserRecommendation.php
foglcz: agreed :+1:

Issue #63, opened 2018-02-20
Support for more combinations of user ids in recommendation request


kubasimon:
Current implementation of recommendation endpoint is checking user ids for (interaction, merge, recommendation)
``` 
     (A,  null,  A)
     (A, A -> ?, ?)
     (?, ? -> A, A)
```

After implementing RAD-1015, client should also allow:

```
   (B, A -> B, B)
```

OndraM:
:+1: Sure, the client will follow the API behavior. Do you know any ETA about the implementation in Matej (RAD-1015)?

OndraM:
This is being rolled now to the API. 

@tenerd Do you need to use the feature now? We may implement it, but it may take a while until we get to it. If you'd like to contribute, it may be sooner - I'll try to do CR ASAP if you will send a PR. I believe the change should not be that big even with tests.

PR #68, opened 2018-04-11
Allow schema (B, A -> B, B) for recommendation request

kubasimon:
(interaction user is same as target user on merge)

Addressing #63

### DIFF
--- a/src/RequestBuilder/RecommendationRequestBuilder.php
+++ b/src/RequestBuilder/RecommendationRequestBuilder.php
@@ -61,8 +61,10 @@ class RecommendationRequestBuilder extends AbstractRequestBuilder
 
     /**
      * Assert that interaction user ids are ok:
+     * ([interaction], [user merge (source -> target)], [recommendation]):
      * - (A,  null,  A)
      * - (A, A -> ?, ?)
+     * - (B, A -> B, B)
      */
     private function assertInteractionUserId(): void
     {
@@ -75,6 +77,13 @@ class RecommendationRequestBuilder extends AbstractRequestBuilder
         // (A, null, A)
         if ($this->userMergeCommand === null && $interactionUserId !== $this->userRecommendationCommand->getUserId()) {
             throw LogicException::forInconsistentUserId($this->userRecommendationCommand, $this->interactionCommand);
+        }
+
+        // allow (B, A -> B, B)
+        if ($this->userMergeCommand !== null
+            && $interactionUserId === $this->userMergeCommand->getUserId()
+            && $interactionUserId === $this->userRecommendationCommand->getUserId()) {
+            return;
         }
 
         // (A, A -> ?, ?)
--- a/tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
+++ b/tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
@@ -113,12 +113,21 @@ class RecommendationRequestBuilderTest extends TestCase
      * ([interaction], [user merge], [recommendation]): (A, A -> B, B)
      *
      * @test
+     * @dataProvider provideCorrectSequenceOfUsers
+     * @param string $interactionUser
+     * @param string $sourceUserToBeDeleted
+     * @param string $targetUserId
+     * @param string $recommendationUser
      */
-    public function shouldPassOnCorrectSequenceOfUsersWhenMerging(): void
-    {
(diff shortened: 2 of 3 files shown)


tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -127,15 +127,33 @@ public function shouldPassOnCorrectSequenceOfUsersWhenMerging(): void
     }
 
     /**
-     * ([interaction], [user merge], [recommendation]): (A, B -> A, A)
+     * ([interaction], [user merge], [recommendation]): (B, A -> B, B)
+     *
+     * @test
+     */
+    public function shouldPassOnCorrectSequenceOfUsersWhenMergingWithInteractionUserSameAsRecommendation(): void
```
OndraM: Maybe there is opportunity for data provider ($interaction, $userMerge, $recommendation) and just extend the previous testcase, what do you think?

src/RequestBuilder/RecommendationRequestBuilder.php
```diff
@@ -72,6 +74,13 @@ private function assertInteractionUserId(): void
 
         $interactionUserId = $this->interactionCommand->getUserId();
 
+        // allow (B, A -> B, B)
```
OndraM: Maybe move this after the following condition, so that cases for "not null user-merge" are next to each other?

OndraM (changes requested):

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
kubasimon: ok, changed to use same test with dataprovider

src/RequestBuilder/RecommendationRequestBuilder.php
kubasimon: moved

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -147,4 +156,12 @@ public function shouldFailOnIncorrectSequenceOfUsersWhenMerging(): void
         $builder->setInteraction($interactionCommand);
         $builder->build();
     }
+
+    public function correctSequenceOfUsersProvider()
+    {
+        return [
+          ['test-user-a', 'test-user-a', 'test-user-b', 'test-user-b'], // (A, A -> B, B)
```
OndraM: The comment could instead be used as array key, its then printed in PHPUnit output, like:

```php
'(A, A -> B, B)' => ['test-user-a', 'test-user-a', 'test-user-b', 'test-user-b'],
```

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -147,4 +156,12 @@ public function shouldFailOnIncorrectSequenceOfUsersWhenMerging(): void
         $builder->setInteraction($interactionCommand);
         $builder->build();
     }
+
+    public function correctSequenceOfUsersProvider()
```
OndraM: - could you please add return typehint?
- also please add phpdoc to help phpstorm & phpstan recognize internal type of array contents
- naming convence for provider method name is "provideFooBar", eg. to not conincidentally name them like "testUsersProvider", which would cause PHpUnit to recognize them as tests and do unexpected things

So this should be:
```php
     /**
     * @return array[]
     */
    public function providecorrectSequenceOfUsers(): array
```

OndraM (changes requested):
Thanks, just to really minor things. I also asked @foglcz to have a quick look as well, but I hope we can merge this then soon - meaning today :).

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
kubasimon: done

tests/unit/RequestBuilder/RecommendationRequestBuilderTest.php
kubasimon: done

OndraM (approved):

OndraM:
Thanks @tenerd ! Let's wait for @foglcz and I will merge it then if he finds no issues :)

foglcz (approved):

PR #71, opened 2018-04-13
should throw assertion exception when calling setCount with non integer parameter

dudla:
for PHP5 version where type don't force integer value

### DIFF
--- a/src/Model/Command/UserRecommendation.php
+++ b/src/Model/Command/UserRecommendation.php
@@ -135,6 +135,7 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
 
     protected function setCount(int $count): void
     {
+        Assertion::integer($count); // for PHP5 clients
         Assertion::greaterThan($count, 0);
 
         $this->count = $count;


kubasimon (approved):
We wasted some time on this issue too 💩

OndraM:
Hi, I understand the reason, but in my opinion, this is simply the tradeoff for using PHP 5 version with weak types. Obvious solution of this issue is to use PHP 7 version - otherwise if we would need type safety in PHP 5, we would need to add assertions like this on tens of places inside the library, where we depend on strict types.

What was the scenario when you came across this issue?

BTW when do you plan to upgrade to PHP 7? I'm thinking about how long could this still be an issue for anyone :).

OndraM:
BTW simpler and for me preferred option would be just to add phpdoc with types. Unfortunately we didn't find an automated (and worrking) solution how to add phpdoc to PHP5 according to the types from PHP7. But if we add them here manually, then if you use eg. PHPStan, it would IMO warn you about incompatible types, maybe PhpStorm will help you as well.

Would this be an solution for you?

dudla:
Sure... with php 7 it will be no problem... this is for php5 version. This is for the time before everybody upgrades. It will save some considerable time debuging for new implementors - and that is the main purpose of the library, isn't it?

In our case it was that a URL parameter was (in some case) passed to setCount without casting to (int). As the default value (used in most cases) was correct *int* constant it was harder to find the origin of the problem.

Other possible solution would be to cast it to (int) on the assignment inside setClient method but it would hide the problem for the client - this way it will report it so that implementor can rather fix their code.

OndraM:
Thinks for the suggestion. We decided to add PHP5 typehins only to PHP5 version via patch applied during transpiliation, to not put this into PHP 7 codebase. See https://github.com/lmc-eu/matej-client-php5/pull/3 .

PR #72, opened 2018-04-13
Fix failing intergration test (RAD-1066)

foglcz:
|  |  |
| --- | --- |
| **Type** | Bugfix |
| **Fixes issues** | n/a |
| **Documentation** | n/a |
| **BC Break** | no |
| **Tests updated** | yes |

Fixes failing integration test, which was failing because of uninitialized item properties.

### DIFF
--- a/tests/integration/IntegrationTestCase.php
+++ b/tests/integration/IntegrationTestCase.php
@@ -8,20 +8,20 @@ use PHPUnit\Framework\TestCase;
 
 class IntegrationTestCase extends TestCase
 {
-    protected function markAsSkippedIfMatejIsNotAvailable(): void
+    protected static function markAsSkippedIfMatejIsNotAvailable(): void
     {
         if (!getenv('MATEJ_TEST_ACCOUNTID')) {
-            $this->markTestSkipped('Environment variable MATEJ_TEST_ACCOUNTID has to be defined');
+            self::markTestSkipped('Environment variable MATEJ_TEST_ACCOUNTID has to be defined');
         }
 
         if (!getenv('MATEJ_TEST_APIKEY')) {
-            $this->markTestSkipped('Environment variable MATEJ_TEST_APIKEY has to be defined');
+            self::markTestSkipped('Environment variable MATEJ_TEST_APIKEY has to be defined');
         }
     }
 
-    protected function createMatejInstance(): Matej
+    protected static function createMatejInstance(): Matej
     {
-        $this->markAsSkippedIfMatejIsNotAvailable();
+        self::markAsSkippedIfMatejIsNotAvailable();
 
         $instance = new Matej(getenv('MATEJ_TEST_ACCOUNTID'), getenv('MATEJ_TEST_APIKEY'));
 
@@ -46,7 +46,7 @@ class IntegrationTestCase extends TestCase
     }
 
     /** @return string[] */
-    protected function generateOkStatuses(int $amount): array
+    protected static function generateOkStatuses(int $amount): array
     {
         $data = explode(',', str_repeat('OK,', $amount));
         array_pop($data);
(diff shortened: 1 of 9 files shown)


tests/integration/RequestBuilder/EventsRequestBuilderTest.php
```diff
@@ -6,13 +6,29 @@
...
+use Lmc\Matej\Model\Command\ItemPropertySetup;
 use Lmc\Matej\Model\Command\UserMerge;
+use Lmc\Matej\RequestBuilder\ItemPropertiesSetupRequestBuilder;
 
 /**
  * @covers \Lmc\Matej\RequestBuilder\EventsRequestBuilder
  */
 class EventsRequestBuilderTest extends IntegrationTestCase
 {
+    protected function setup(): void
```
OndraM: This should be setUp, and I'm wondering why this was not detected by codestyle check, there should be one exactly for this 🤔

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
```diff
@@ -6,13 +6,29 @@
...
+    {
+        $builder = $this->createMatejInstance()->request()->setupItemProperties();
+
+        $this->buildAndSendPropertySetupRequest($builder);
```
OndraM: Is it OK this will be done before each test method in this testcase?

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
```diff
@@ -51,4 +67,12 @@ public function shouldExecuteInteractionAndUserMergeAndItemPropertyCommands(): v
 
         $this->assertResponseCommandStatuses($response, ...$this->generateOkStatuses(10));
     }
+
+    private function buildAndSendPropertySetupRequest(ItemPropertiesSetupRequestBuilder $builder): void
+    {
+        $builder->addProperty(ItemPropertySetup::string('test_property_a'))
```
OndraM: We may add the identifier to some private class constants, to signalize these are "predefined" item properties available for use in this test. What do you thing?

OndraM (changes requested):

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
OndraM: https://github.com/lmc-eu/php-coding-standard/issues/14

tests/integration/IntegrationTestCase.php
```diff
@@ -32,21 +32,21 @@ protected function createMatejInstance(): Matej
         return $instance;
     }
 
-    protected function assertResponseCommandStatuses(Response $response, ...$expectedCommandStatuses): void
+    protected static function assertResponseCommandStatuses(Response $response, ...$expectedCommandStatuses): void
```
OndraM: Why is this static? This should be non-static, co it could be used in a same way as other phpunit asserts (via $this->assertResponseCommandStatuses, not static::assertResponseCommandStatuses)

tests/integration/IntegrationTestCase.php
```diff
@@ -8,20 +8,20 @@
...
         }
 
         if (!getenv('MATEJ_TEST_APIKEY')) {
-            $this->markTestSkipped('Environment variable MATEJ_TEST_APIKEY has to be defined');
+            self::markTestSkipped('Environment variable MATEJ_TEST_APIKEY has to be defined');
         }
     }
 
-    protected function createMatejInstance(): Matej
+    protected static function createMatejInstance(): Matej
```
OndraM, line 22: If you declare the method static, you should use it via static call eveywhere where it is being used (in all other testcases)... Like `static::createMatejInstance()`

OndraM (changes requested):

tests/integration/IntegrationTestCase.php
OndraM: This is what happens then:
static

So if we don't need to, lets keep this non-static :)

OndraM (changes requested):

tests/integration/IntegrationTestCase.php
foglcz: :+1: fixed, thanks :)

OndraM (approved):

OndraM:
Nice job @foglcz, thanks! :+1:

PR #73, opened 2018-04-13
Implement A/B testing support for Recommendations and Sorting commands

foglcz:
|  |  |
| --- | --- |
| **Type** | Feature |
| **Fixes issues** | RAD-1066 |
| **BC Break** | no |
| **Tests updated** | yes |

Contains implementation for A/B testing of various recommendation models. These models have to be implemented/deployed in Matej first, before they're available for any given client. This is by design, as new model requires full recalculation, and therefore availability is individual depending on the size of clients data.

By default, only `default` model is available for every client (which means *clients* default, not *matej* default.)

Requires **Matej release v7.36**
Currently deployed to Staging instance, which contains models `default` and `model_b`.

There is no API endpoint that'd list available models. No validation beyond basic identifier is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,51 @@ $response = $matej->request()
     ->send();
 ```
 
+### A/B Testing support
+`Recommendation` and `Sorting` commands support optional A/B testing of various models. This has to be set up in Matej first,
+but once available, you can specify which model you want to use when requesting recommendations or sortings.
+
+This is available for `recommendation`, `sorting` and `campaign` requests:
+
+```php
+$recommendationCommand = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
+$recommendationCommand->setModelName('alpha');
+
+$sortingCommand = Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3']);
+$sortingCommand->setModelName('beta')
+
+$response = $matej->request()
+    ->recommendation($recommendationCommand)
+    ->send();
+
+$response = $matej->request()
+    ->sorting($sortingCommand)
+    ->send();
+
+$response = $matej->request()
+    ->campaign()
+    ->addRecommendation($recommendationCommand->setModelName('gamma'))
+    ->addSorting($sortingCommand->setModelName('delta'))
+    ->send();
+```
+
+If you don't provide any model name, the request will be sent without it, and Matej will use default model for your instance.
+
+Typically, you'd select a random sample of users, to which you'd present recommendations and sortings from second model. This way, implementation
+in your code should look similar to this:
+
+```php
+$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
+
+if ($session->isUserInBucketB()) {
+    $recommendation->setModelName('alpha');
+}
+
+$response = $matej->request()->recommendation($recommendation)->send();
+```
+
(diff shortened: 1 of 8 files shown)


src/Model/Command/Sorting.php
```diff
@@ -31,6 +33,20 @@ public static function create(string $userId, array $itemIds): self
...
+     *
+     * @return $this
+     */
+    public function setModelName(string $modelName): self
+    {
+        Assertion::typeIdentifier($modelName);
+
+        $this->modelName = $modelName;
+
+        return $this;
```
OndraM, line 47: We don't use fluent interfaces in this kind of setters.

README.md
```diff
@@ -251,6 +251,45 @@ $response = $matej->request()
...
+    ->send();
+
+$response =  $matej->request()
+    ->campaign()
+    ->addRecommendation(UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600)->setModelName('gamma'))
+    ->addSorting(Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3'])->setModelName('delta'))
+    ->send();
+```
+
+If you don't provide any model name, the request will be sent without it, and matej will use default model for your instance.
```
OndraM: *M*atej

README.md
```diff
@@ -251,6 +251,45 @@ $response = $matej->request()
...
+but once available, you can specify which model you want to use when requesting recommendations or sortings.
+
+This is available for `recommendation`, `sorting` and `campaign` requests:
+
+```php
+$response = $matej->request()
+    ->recommendation(UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600)->setModelName('alpha'))
+    ->send();
+
+$response =  $matej->request()
```
OndraM: Two spaces before $matej (also five lines below)

OndraM (changes requested):

README.md
```diff
@@ -251,6 +251,45 @@ $response = $matej->request()
...
 ```
 
+### A/B Testing support
+`Recommendation` and `Sorting` commands support optional A/B testing of various models. This has to be set up in Matej first,
+but once available, you can specify which model you want to use when requesting recommendations or sortings.
+
+This is available for `recommendation`, `sorting` and `campaign` requests:
+
+```php
```
OndraM: You can try separating the commands to standalone variable - it could be more readable in comparison with those "oneliners".

src/Model/Command/Sorting.php
foglcz: I thought we do? `Public` setters are generally fluent, `protected` setters are not.

I used `UserRecommendation` as code-style template https://github.com/lmc-eu/matej-client-php/blob/master/src/Model/Command/UserRecommendation.php#L103

README.md
foglcz: Nice catch! :) :+1:

README.md
foglcz: :+1:

README.md
foglcz: :+1:

src/Model/Command/Sorting.php
OndraM: Sorry, my fault, I realized this later, but forgot to delete the comment :see_no_evil:

OndraM (approved):

PR #82, opened 2018-07-24
Implement support for Item Properties in Recommendation Requests

foglcz:
|  |  |
| --- | --- |
| **Type** | Feature |
| **Fixes issues** | RAD-1530 |
| **BC Break** | **YES** |
| **Tests updated** | yes |

Implements new feature in Matej, which allows you to request list of item properties you'd like returned as part of recommendation request. This allows you to request, say, a job position title, location and perex, directly as part of the recommendation request - cleaning up the codebase (because you don't need to do a roundtrip to local db with item ids)

**Requires Matej 7.39**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,14 @@
 <!-- There is always Unreleased section on the top. Subsections (Added, Changed, Fixed, Removed) should be added as needed. -->
 
 ## Unreleased
+### Changed
+- **BC BREAK** | `UserRecommendation` now returns new format of response in `->getData()`, which is a list of `stdClass` instances.
+
 ### Fixed
 - Exceptions occurring during async request now generate rejected promise (as they should) and are no longer thrown directly.
+
+### Added
+- Ability to request item properties in `UserRecommendation` command, which are then returned by Matej alongside with `item_id`.
 
 ## 1.6.0 - 2018-06-01
 ### Added
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ foreach ($response->getCommandResponses() as $commandResponse) {
 and [Item Properties](#item-properties-setup-to-setup-you-matej-database) endpoints have syntax sugar shortcuts,
 which makes processing responses easier. See below for detailed examples.
 
-### Item properties setup (to setup you Matej database)
+### Item properties setup (to setup your Matej's database)
 
 ```php
 $matej = new Matej('accountId', 'apikey');
@@ -205,7 +205,63 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
 echo $response->getRecommendation()->getStatus(); // OK
 
 $recommendations = $response->getRecommendation()->getData();
+
+// var_dump($recommendations):
+// array(2) {
+//     [0] => object(stdClass)#1 (2) {
+//         ["item_id"]  => string(9) "item_id_1"
+//     }
+//     [1] => object(stdClass)#2 (2) {
+//         ["item_id"]  => string(9)  "item_id_2"
+//     }
+// }
 ```
+
+#### Recommendation response properties
+
+Every item in Matej has its id, and optionally other item properties. These properties can be set up in [item properties setup](#item-properties-setup-to-setup-you-matej-database),
+and you can upload item data in the [events](#send-events-data-to-matej) request. This has major benefit because you can request
+these properties to be returned as part of your Recommendation Request.
+
+We call them response properties, and they can be specified either as the last parameter of `UserRecommendation::create` function,
+by calling `->addResponseProperty()` method, or by calling `->setResponseProperties()` method. Following will request an `item_id`,
+`item_url` and `item_title`:
+
+```php
+$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600, ['item_url'])
+    ->addResponseProperty('item_title');
+
+$response = $matej->request()
+    ->recommendation($recommendation)
+    ->send();
+
+$recommendedItems = $response->getRecommendation()->getData();
+
+// $recommendedItems is an array of stdClass instances:
+//
+// array(2) {
+//     [0] => object(stdClass)#1 (2) {
+//         ["item_id"]  => string(9) "item_id_1"
+//         ["item_url"] => string(5) "url_1"
+//         ["item_title"] => string(5) "title_1"
+//     }
+//     [1] => object(stdClass)#2 (2) {
+//         ["item_id"]  => string(9)  "item_id_2"
+//         ["item_url"] => string(10) "url_2"
+//         ["item_title"] => string(10) "title_2"
+//     }
+// }
+```
+
+If you don't specify any response properties, Matej will return an array of `stdClass` instances, which contain only `item_id` property.
+If you do request at least one response property, you don't need to metion `item_id`, as Matej will always return it regardless of the
+properties requested.
+
+If you request an unknown property, Matej will return a `BAD REQUEST` with HTTP status code `400`.
+
+This way, when you receive recommendations from Matej, you don't need to loop the `item_id` and retrieve further information
+from your local database. It means, however, that you'll have to keep all items up to date within Matej,
+which can be done through [events](#send-events-data-to-matej) request.
 
 ### Request item sorting for single user
 
--- a/src/Model/Command/UserRecommendation.php
+++ b/src/Model/Command/UserRecommendation.php
@@ -29,18 +29,27 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
     private $hardRotation = false;
     /** @var string */
     private $minimalRelevance = self::MINIMAL_RELEVANCE_LOW;
-    /** @var array */
+    /** @var string[] */
     private $filters = ['valid_to >= NOW'];
     /** @var string|null */
     private $modelName = null;
+    /** @var string[] */
+    private $responseProperties = [];
 
-    private function __construct(string $userId, int $count, string $scenario, float $rotationRate, int $rotationTime)
-    {
+    private function __construct(
+        string $userId,
+        int $count,
+        string $scenario,
+        float $rotationRate,
+        int $rotationTime,
+        array $responseProperties
+    ) {
         $this->setUserId($userId);
         $this->setCount($count);
         $this->setScenario($scenario);
         $this->setRotationRate($rotationRate);
         $this->setRotationTime($rotationTime);
+        $this->setResponseProperties($responseProperties);
     }
 
     /**
@@ -53,6 +62,7 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
      * Set from 0.0 for no rotation (same items will be recommended) up to 1.0 (same items should not be recommended).
      * @param int $rotationTime Specify for how long will the item's rotationRate be taken in account and so the item
      * is penalized for recommendations.
+     * @param string[] $responseProperties Specify which properties you want to retrieve from Matej alongside the item_id.
      * @return static
      */
     public static function create(
@@ -60,9 +70,10 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
         int $count,
         string $scenario,
         float $rotationRate,
-        int $rotationTime
+        int $rotationTime,
+        array $responseProperties = []
     ): self {
-        return new static($userId, $count, $scenario, $rotationRate, $rotationTime);
+        return new static($userId, $count, $scenario, $rotationRate, $rotationTime, $responseProperties);
     }
 
     /**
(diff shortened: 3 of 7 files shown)


src/Model/Command/UserRecommendation.php
```diff
@@ -123,6 +125,33 @@ public function setFilters(array $filters): self
         return $this;
     }
 
+    /**
+     * Specify which item property you want returned.
+     */
+    public function addProperty(string $property): self
```
OndraM: What about some more descriptive name? `addReturnProperty`  / `addResponseProperty` etc.?

This may appear as settings of the recommendation, not about contents of the response 🤔,

tests/integration/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -29,6 +29,19 @@ public function shouldExecuteRecommendationRequestOnly(): void
         $this->assertShorthandResponse($response, 'SKIPPED', 'SKIPPED', 'OK');
     }
 
+    /** @test */
+    public function shouldExecuteRecommendationRequestWithPropertiesSet(): void
+    {
+        $response = static::createMatejInstance()
+            ->request()
+            ->recommendation($this->createRecommendationCommand('user-a')->setProperties(['item_id']))
```
OndraM: The Recommendation should be for readability instantiated separately

```
$recommendation = $this->createRecommendationCommand('user-a')
    ->setProperties(['item_id']);
...
```

OndraM: The Recommendation should be for readability instantiaded separately

```
$recommendation = $this->createRecommendationCommand('user-a')
    ->setProperties(['item_id']);
...
```

tests/integration/RequestBuilder/RecommendationRequestBuilderTest.php
```diff
@@ -57,6 +70,19 @@ public function shouldFailOnInvalidModelName(): void
...
+    /** @test */
+    public function shouldFailOnInvalidPropertyName(): void
+    {
+        $this->expectException(RequestException::class);
+        $this->expectExceptionCode(400);
+        $this->expectExceptionMessage('BAD REQUEST');
+
+        $this->createMatejInstance()
+            ->request()
+            ->recommendation($this->createRecommendationCommand('user-a')->addProperty('unknown-property'))
```
OndraM: dtto, separate into own property

tests/unit/Http/Fixtures/response-recommendation-v1.json
```diff
@@ -0,0 +1,38 @@
...
+          "jd-5a2ebcb2fc13ae1fdb0002cb"
+        ],
+        "message": "",
+        "status": "OK"
+      }
+    ]
+  },
+  "message": "",
+  "status": "OK"
+}
```
OndraM: missing newline, also in the second json

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 ### Fixed
 - Exceptions occurring during async request now generate rejected promise (as they should) and are no longer thrown directly.
 
+### Added
+- Support for setting `properties` parameter in `UserRecommendation`.
```
OndraM: This is written more from API point of view; what about something like
`- Ability to set properties expected in recommendation response (via add...() or set...())`

README.md
```diff
@@ -201,6 +201,100 @@ echo $response->getRecommendation()->getStatus(); // OK
 $recommendations = $response->getRecommendation()->getData();
 ```
 
+#### Recommendation response format
+
+*New in Matej v7.39*
```
OndraM: This could be omitted, if we merge this only after 7.39 is on production :)

OndraM: Also, as we discussed personally, we may target this to v2.0 of the client, and transparently declare change of the Recommendation format (from array of ids to array of objects with multiple properties) as a BC break.

Thoughts from others? @jirinovak @tenerd etc. ?

src/Model/Command/UserRecommendation.php
```diff
@@ -29,10 +29,12 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
     private $hardRotation = false;
     /** @var string */
     private $minimalRelevance = self::MINIMAL_RELEVANCE_LOW;
-    /** @var array */
+    /** @var string[] */
     private $filters = ['valid_to >= NOW'];
     /** @var string|null */
     private $modelName = null;
+    /** @var string[] */
+    private $properties = null; // @v2.0 use ['item_id'] as default
```
OndraM: If `['item_id'] ` will be default and required in v2.0, should you be able to reset it (so also remove the item_id and possibly trigger the old format) using `setProperties()`? I think you shouldn't... So maybe add the `['item_id']` always, and use this array only for additional properties?

OndraM (changes requested):

OndraM:
Also please rebase this with master to fix codestyle check :)

foglcz:
@OndraM Fixed review, updated as per discussion to be a bc-break targetting sdk 2.0

README.md
```diff
@@ -199,8 +199,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+//     }
+//     [1] => object(stdClass)#2 (2) {
+//         ["item_id"]  => string(9)  "item_id_2"
+//     }
+// }
 ```
 
+#### Recommendation response properties
+
+Every item in matej has it's id, and optionally other item properties. These properties can be set up in [item properties setup](#item-properties-setup-to-setup-you-matej-database),
```
OndraM: its

README.md
```diff
@@ -205,8 +205,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+and you can upload item data in the [events](#send-events-data-to-matej) request. This has major benefit, because you can request
```
OndraM: unnecessary comma after "benefit"

README.md
OndraM: *M*atej :)?

README.md
```diff
@@ -205,8 +205,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+these properties to be returned as part of your Recommendation Request.
+
+We call them response properties, and they can be specified either as last parameter of `UserRecommendation::create` function,
```
OndraM: as the last

README.md
```diff
@@ -205,8 +205,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+by calling `->addResponseProperty()` method, or by calling `->setResponseProperties()` method. Following will request an `item_id`,
+`item_url` and `item_title`:
+
+```php
+$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600, ['item_id', 'item_url')
+    ->addProperty('item_title')
```
OndraM: semicolon could be on the same line

README.md
```diff
@@ -205,8 +205,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+Every item in matej has it's id, and optionally other item properties. These properties can be set up in [item properties setup](#item-properties-setup-to-setup-you-matej-database),
+and you can upload item data in the [events](#send-events-data-to-matej) request. This has major benefit, because you can request
+these properties to be returned as part of your Recommendation Request.
+
+We call them response properties, and they can be specified either as last parameter of `UserRecommendation::create` function,
+by calling `->addResponseProperty()` method, or by calling `->setResponseProperties()` method. Following will request an `item_id`,
+`item_url` and `item_title`:
+
+```php
+$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600, ['item_id', 'item_url')
```
OndraM: missing `]`

README.md
```diff
@@ -205,8 +205,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+//     }
+//     [1] => object(stdClass)#2 (2) {
+//         ["item_id"]  => string(9)  "item_id_2"
+//         ["item_url"] => string(10) "url_2"
+//         ["item_title"] => string(10) "title_2"
+//     }
+// }
+```
+
+Matej will always return `item_id`, even if you don't ask for it explicitly.
```
OndraM: So, what is the point of asking for it explicitly :)?

Should be the best practice to ask for all fields, or to just list additional fields (beyond item_id)?

Because in phpdoc of UserRecommendation::create, you advice "which properties you want to retrieve from matej alongside the item_id" - but example above includes item_id. This should be unified to adhere to the best practice, whatever it is.

src/Model/Command/UserRecommendation.php
```diff
@@ -204,6 +242,10 @@ protected function getCommandParameters(): array
             $parameters['model_name'] = $this->modelName;
         }
 
+        if ($this->responseProperties !== null) {
```
OndraM: So you don't send item_id, and it is always returned. In this case I would omit item_id from the examples above and just explain it is always included, and you define just additional properties.

tests/unit/Http/ResponseDecoderTest.php
```diff
@@ -96,4 +97,34 @@ public function shouldThrowExceptionWhenJsonWithInvalidDataIsDecoded(): void
         $this->expectExceptionMessage('"invalid": [],');
         $this->decoder->decode($response);
     }
+
+    /** @test */
+    public function shouldDecodeVersionOneRecommendationResponse(): void
```
OndraM: So is there backward compatibility or not? I thought once you use library with this updated code, you will get response in the new format (that is why it is BC break, right?). If you could still get array of string from `getData`, how could you safely update you app for the new format?

OndraM (changes requested):

src/Model/Command/UserRecommendation.php
```diff
@@ -53,16 +62,18 @@ private function __construct(string $userId, int $count, string $scenario, float
      * Set from 0.0 for no rotation (same items will be recommended) up to 1.0 (same items should not be recommended).
      * @param int $rotationTime Specify for how long will the item's rotationRate be taken in account and so the item
      * is penalized for recommendations.
+     * @param string[] $responseProperties Specify which properties you want to retrieve from matej alongside the item_id.
```
OndraM: *M*atej

README.md
```diff
@@ -205,8 +205,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+
+This way, when you receive recommendations from Matej, you don't need to loop the `item_id` and retrieve further information
+from your local database. It means however, that you'll have to keep all items up to date within Matej,
```
OndraM: Either
> It means, however, that 

or maybe

> It means however that 

:)

tests/unit/Http/ResponseDecoderTest.php
foglcz: You have a point here; there is a backwards compatibility but we don't want to expose it in the PHP client, as it will get removed from Matej at some point. I've updated the Client, so that it's a clean BC-break.

src/Model/Command/UserRecommendation.php
```diff
@@ -123,6 +134,33 @@ public function setFilters(array $filters): self
...
+    {
+        Assertion::typeIdentifier($property);
+
+        $this->responseProperties[] = $property;
+
+        return $this;
+    }
+
+    /**
+     * Overwrite all properties by custom specified list. Note that this will overried the defaults.
```
OndraM: > Note that this will overried the defaults.

Which defaults? By default there are none.

README.md
```diff
@@ -205,8 +205,60 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+and you can upload item data in the [events](#send-events-data-to-matej) request. This has major benefit because you can request
+these properties to be returned as part of your Recommendation Request.
+
+We call them response properties, and they can be specified either as the last parameter of `UserRecommendation::create` function,
+by calling `->addResponseProperty()` method, or by calling `->setResponseProperties()` method. Following will request an `item_id`,
+`item_url` and `item_title`:
+
+```php
+$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600, ['item_id', 'item_url'])
+    ->addProperty('item_title');
```
OndraM: addResponseProperty

tests/unit/Http/ResponseDecoderTest.php
OndraM: So will you drop this testcase and also support for the old format :)? (and also response-recommendation-v1.json fixture )

README.md
```diff
@@ -205,8 +205,60 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+//     }
+//     [1] => object(stdClass)#2 (2) {
+//         ["item_id"]  => string(9)  "item_id_2"
+//         ["item_url"] => string(10) "url_2"
+//         ["item_title"] => string(10) "title_2"
+//     }
+// }
 ```
 
+Matej will always return `item_id`, even if you don't ask for it.
```
OndraM: See https://github.com/lmc-eu/matej-client-php/pull/82#discussion_r213621372 - still valid :)

OndraM (changes requested):

CHANGELOG.md
```diff
@@ -5,9 +5,15 @@
...
 
 ## Unreleased
+### Changed
+- **BC BREAK** | `UserRecommendation` now returns new format of response in `->getData()`, which is a list of `stdClass` instances.
+
 ### Fixed
 - Exceptions occurring during async request now generate rejected promise (as they should) and are no longer thrown directly.
 
+### Added
+- `UserRecommendation` now sends which item properties should be returned alongside with item_ids.
```
OndraM: Maybe something `like

     - Possibility to define in `UserRecommendation` command which item properties should be returned alongside with item_ids.

OndraM (changes requested):
Just the changelog entry could be more clear, otherwise :+1: from me :)

OndraM (approved):

jirinovak (approved):

OndraM:
Thanks @foglcz!

PR #83, opened 2018-08-28
Add MQL support for recommendations filtering

foglcz:
|  |  |
| --- | --- |
| **Type** | Feature |
| **Fixes issues** | RAD-1689 |
| **BC Break** | **YES** |
| **Tests updated** | yes |

Adds support for Matej Query Language in the recommendation requests, which is now default way of sending requests.

Removes default filter.

Legacy filtering using regexes can be done by setting filter type to `FILTER_TYPE_RGX`. In the request itself, the legacy way of filtering is implemented by _absence_ of parameter `filter_type`. However, we don't want to encourage Matej users to use the RGX filter, as MQL is overall faster, *and* implements more features. Therefore, I did not mention this explicitly in the documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,16 @@
 <!-- There is always Unreleased section on the top. Subsections (Added, Changed, Fixed, Removed) should be added as needed. -->
 
 ## Unreleased
+### Changed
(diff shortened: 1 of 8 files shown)


foglcz:
This issue has been closed because I've mistakenly squashed wrong commits. Sorry!

PR #84, opened 2018-08-28
Change filtering language to MQL

foglcz:
|  |  |
| --- | --- |
| **Type** | Feature |
| **Fixes issues** | RAD-1689 |
| **BC Break** | **YES** |
| **Tests updated** | yes |
| **Notes** | Requires #82 to be merged first. |

Adds support for Matej Query Language in the recommendation requests, which is now default way of sending requests.

Removes default filter.

Legacy filtering using regexes can be done by setting filter type to `FILTER_TYPE_RGX`. In the request itself, the legacy way of filtering is implemented by _absence_ of parameter `filter_type`. However, we don't want to encourage Matej users to use the RGX filter, as MQL is overall faster, *and* implements more features. Therefore, I did not mention this explicitly in the documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ## Unreleased
 ### Changed
 - **BC BREAK** | `UserRecommendation` now returns new format of response in `->getData()`, which is a list of `stdClass` instances.
+- **BC BREAK** | `UserRecommendation` does not have default filter (was previously set to: `valid_to >= NOW`).
+- **BC BREAK** | `UserRecommendation` now uses MQL query language by default for filtering.
 
 ### Fixed
 - Exceptions occurring during async request now generate rejected promise (as they should) and are no longer thrown directly.
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ You can also set more granular options of the recommendation command:
 
 ```php
 $recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
-$recommendation->setFilters(['valid_to >= NOW']) // Note this filter is present by default
+$recommendation->setFilters(['for_recommendation = 1'])
     ->setMinimalRelevance(UserRecommendation::MINIMAL_RELEVANCE_HIGH)
     ->enableHardRotation();
 
--- a/src/Model/Command/UserRecommendation.php
+++ b/src/Model/Command/UserRecommendation.php
@@ -12,6 +12,7 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
     public const MINIMAL_RELEVANCE_LOW = 'low';
     public const MINIMAL_RELEVANCE_MEDIUM = 'medium';
     public const MINIMAL_RELEVANCE_HIGH = 'high';
+    public const FILTER_TYPE_MQL = 'mql';
 
     /** @var string */
     protected $filterOperator = 'and';
@@ -30,7 +31,9 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
     /** @var string */
     private $minimalRelevance = self::MINIMAL_RELEVANCE_LOW;
     /** @var string[] */
-    private $filters = ['valid_to >= NOW'];
+    private $filters = [];
+    /** @var string */
+    private $filterType = self::FILTER_TYPE_MQL;
     /** @var string|null */
     private $modelName = null;
     /** @var string[] */
@@ -236,6 +239,7 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
             'hard_rotation' => $this->hardRotation,
             'min_relevance' => $this->minimalRelevance,
             'filter' => $this->assembleFiltersString(),
+            'filter_type' => $this->filterType,
             'properties' => $this->responseProperties,
         ];
 
--- a/tests/unit/Model/Command/UserRecommendationTest.php
+++ b/tests/unit/Model/Command/UserRecommendationTest.php
@@ -23,9 +23,10 @@ class UserRecommendationTest extends TestCase
                     'rotation_time' => 3600,
                     'hard_rotation' => false,
                     'min_relevance' => UserRecommendation::MINIMAL_RELEVANCE_LOW,
-                    'filter' => 'valid_to >= NOW',
+                    'filter' => '',
+                    'filter_type' => UserRecommendation::FILTER_TYPE_MQL,
                     'properties' => [],
-                    // intentionally no model name ==> should be absent when not used
+                    // when using default model name, parameter "model_name" should be absent.
                 ],
             ],
             $command->jsonSerialize()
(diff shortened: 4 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,9 +5,17 @@
...
+### Changed
+- **BC BREAK** | `UserRecommendation` now returns new format of response in `->getData()`, which is a list of `stdClass` instances.
+- **BC BREAK** | `UserRecommendation` does not have default filter (was previously set to: `valid_to >= NOW`).
+
 ### Fixed
 - Exceptions occurring during async request now generate rejected promise (as they should) and are no longer thrown directly.
 
+### Added
+- `UserRecommendation` now sends which item properties should be returned alongside with item_ids.
+- **BC BREAK** | `UserRecommendation` now uses MQL query language by default for filtering.
```
jirinovak: Maybe this should be in `Changed` section.

README.md
```diff
@@ -205,8 +205,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+and you can upload item data in the [events](#send-events-data-to-matej) request. This has major benefit, because you can request
+these properties to be returned as part of your Recommendation Request.
+
+We call them response properties, and they can be specified either as last parameter of `UserRecommendation::create` function,
+by calling `->addResponseProperty()` method, or by calling `->setResponseProperties()` method. Following will request an `item_id`,
+`item_url` and `item_title`:
+
+```php
+$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600, ['item_id', 'item_url')
+    ->addProperty('item_title')
```
jirinovak: typo `addResponseProperty` according to text above

README.md
```diff
@@ -205,8 +205,61 @@ echo $response->getUserMerge()->getStatus();      // SKIPPED
...
+//     }
+//     [1] => object(stdClass)#2 (2) {
+//         ["item_id"]  => string(9)  "item_id_2"
+//         ["item_url"] => string(10) "url_2"
+//         ["item_title"] => string(10) "title_2"
+//     }
+// }
+```
+
+Matej will always return `item_id`, even if you don't ask for it explicitly.
```
jirinovak: What will be return when i do not request any response properties? Only `item_id` or all item properties?

jirinovak (changes requested):

tests/unit/Model/Command/UserRecommendationTest.php
```diff
@@ -63,6 +65,7 @@ public function shouldUseCustomParameters(): void
                     'hard_rotation' => true,
                     'min_relevance' => UserRecommendation::MINIMAL_RELEVANCE_HIGH,
                     'filter' => 'foo = bar and baz = ban',
+                    // when a filter type is RGX, the parameter is absent.
```
OndraM: which parameter?

(I know which, but its not obvious from the comment :) )

src/Model/Command/UserRecommendation.php
```diff
@@ -134,6 +138,18 @@ public function setFilters(array $filters): self
         return $this;
     }
 
+    /**
+     * Specify the filter type being used.
+     */
+    public function setFilterType(string $filterType): self
+    {
+        Assertion::typeIdentifier($filterType);
```
OndraM: Also assert the allowed values:
```
Assertion::choice(
    $filterType,
    [static::FILTER_TYPE_RGX, static::FILTER_TYPE_MQL]
);
```

(like we do in `setMinimalRelevance()` etc.)

src/Model/Command/UserRecommendation.php
```diff
@@ -30,7 +32,9 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
     /** @var string */
     private $minimalRelevance = self::MINIMAL_RELEVANCE_LOW;
     /** @var string[] */
-    private $filters = ['valid_to >= NOW'];
+    private $filters = [];
```
OndraM, line 34: Also phpdoc of `addFilter()` and `setFilters()` method needs to be updated to reflect this change (= there is no default filter).

tests/unit/Model/Command/UserRecommendationTest.php
```diff
@@ -77,14 +80,14 @@ public function shouldAssembleFilters(): void
         $command = UserRecommendation::create('user-id', 333, 'test-scenario', 1.0, 3600);
 
         // Default filter
-        $this->assertSame('valid_to >= NOW', $command->jsonSerialize()['parameters']['filter']);
+        $this->assertSame('', $command->jsonSerialize()['parameters']['filter']);
 
         // Add custom filters to the default one
         $command->addFilter('foo = bar')
```
OndraM: This test method tests just the legacy regexp filters, right?

I'd thus rename it (`shouldAssembleLegacyRgxFilters()` or something like this) and add another one using some basic MQL filters. Test serve as documentation, and right now, there is no testcase actually using MQL filters.

OndraM (changes requested):
I'm missing some examples and documentation of the filters in README, do you plan to add it? (or later?)

README.md
foglcz: Matej will return only `item_id` - but in the new format of `array of stdClass`-es. 

I've updated the doc to be more explicit.

README.md
foglcz: Good catch! thanks :)

foglcz:
Updated the PR as per rebiew remarks.
ping @OndraM @jirinovak

tests/unit/Model/Command/UserRecommendationTest.php
```diff
@@ -94,6 +97,38 @@ public function shouldAssembleFilters(): void
...
+            ->addFilter('nullable_property IS NULL')
+            ->addFilter("'some_value' in set_property");
+
+        $this->assertSame(
+            "first_string_property = 'bar' and " .
+                "second_string_property LIKE 'bar%' and " .
+                "third_string_property NOT LIKE '%bar' and " .
+                'bool_property = true and ' .
+                'nullable_property IS NULL and ' .
+                "'some_value' in set_property",
```
jirinovak: Maybe NOWDOC would be more readable.

Or at least i would align `"`.

jirinovak (approved):

src/Model/Command/UserRecommendation.php
```diff
@@ -134,6 +137,22 @@ public function setFilters(array $filters): self
         return $this;
     }
 
+    /**
+     * Specify the filter type being used.
+     */
+    public function setFilterType(string $filterType): self
```
OndraM: This whole method seems uncovered by unit tests. I think there should be (simple) testcase for the rgx type.

OndraM (changes requested):

src/Model/Command/UserRecommendation.php
foglcz: After discussion, we've removed this one :)

OndraM (approved):

OndraM:
Marged to master, thanks @foglcz :100:

PR #87, opened 2018-09-13
Add support for allow_seen in Recommendation requests

foglcz:
|  |  |
| --- | --- |
| **Type** | Feature |
| **Fixes issues** | RAD-1735 |
| **Requires Matej** | `>= 7.46.0` |
| **BC Break** | No |
| **Tests updated** | yes |
| **Notes** | n/a |

By default, "seen" is any item, that the user has DetailView'ed. This however, can be changed for any given model, so "seen" can mean different things in `default` and `gamma` model for any particular client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ## Unreleased
 
+### Added
+- `UserRecommendation` has new method `setAllowSeen(true|false)`, which allows recommending of already visited items.
+
 ## 2.0.0 - 2018-09-04
 See [UPGRADE-2.0.md](UPGRADE-2.0.md) for upgrade instructions.
 
--- a/src/Model/Command/UserRecommendation.php
+++ b/src/Model/Command/UserRecommendation.php
@@ -36,6 +36,8 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
     private $modelName;
     /** @var string[] */
     private $responseProperties = [];
+    /** @var bool */
+    private $allowSeen = false;
 
     private function __construct(
         string $userId,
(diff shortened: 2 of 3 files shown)


VincTheSecond (dismissed):

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ## Unreleased
 
+### Added
+- `UserRecommendation` can send `allow_seen`, which allows recommending of already visited items.
```
OndraM: This (`allow_seen`) is from API point of view, maybe rather mention `setAllowSeen()` method from the API Client pow

src/Model/Command/UserRecommendation.php
```diff
@@ -173,6 +175,20 @@ public function setModelName(string $modelName): self
         return $this;
     }
 
+    /**
+     * Allow items, that the user has already "seen"
+     *
+     * By default user won't see any items, that he or she has visitted (and we have recorded DetailView interaction.)
```
OndraM: `it has visited`

src/Model/Command/UserRecommendation.php
```diff
@@ -173,6 +175,20 @@ public function setModelName(string $modelName): self
...
+     * If you want to circumvent this, and get recommendations including the ones, that the user has already visitted,
+     * you can set the "seen" allowance here.
+     */
+    public function setAllowSeen(bool $seen = true): self
```
OndraM: This is confusing - we should either has setter `setAllowSeen(bool)` with **required** parameter, or two methods to set it to true and false (`allowSeen()` and eg. `disallowSeen()`). For simplicity I'd go for the first variant, but for sure not mix those two approaches.

src/Model/Command/UserRecommendation.php
```diff
@@ -242,6 +258,10 @@ protected function getCommandParameters(): array
             $parameters['model_name'] = $this->modelName;
         }
 
+        if ($this->allowSeen) {
```
OndraM: Rather strict comparison (eg. `!== false`) instead of this truthy condition.

OndraM (changes requested):

OndraM (approved):

PR #91, opened 2018-10-02
Add release how-to

OndraM:
Ref. AR-129

This is to help others follow proper release process, in case of me being away 🏝 🍹.

### DIFF
--- a/RELEASE.MD
+++ b/RELEASE.MD
@@ -1,0 +1,28 @@
+# Releasing new Matej PHP Client version
+
+1. Check [open pull-requests](https://github.com/lmc-eu/matej-client-php/pulls), if there are any, you may coordinate releasing of a new version with authors of these pull requests.
+1. If there are some backward incompatible changes, the specific message in CHANGELOG.md must be prefixed with `**BC BREAK**`.
+1. Now new version number needs to be determined. The format is `MAJOR.MINOR.PATCH` and you MUST follow [SemVer rules](http://semver.org/):
+    - Increase MAJOR version number, if there are any API backward incompatible changes, eg. renaming/removing of any public interface etc.
+    - Increase MINOR version number when you add functionality in a backward-compatible manner.
+    - Increase PATCH version if the release contains *only* backward-compatible bug fixes.
+1. Update [CHANGELOG.md](CHANGELOG.md):
+    - Add header in form `## X.Y.Z - YYYY-MM-DD` above the list of changes that were already merged.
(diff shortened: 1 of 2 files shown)


legendik (approved):

RELEASE.MD
```diff
@@ -0,0 +1,28 @@
+# Releasing new Matej PHP Client version
+
+1. Check [open pull-requests](https://github.com/lmc-eu/matej-client-php/pulls), if there are any, you may coordinate releasing of a new version with authors of these pull requests.
```
MortalFlesh, line 3: why are there only `1.` ?

RELEASE.MD
```diff
@@ -0,0 +1,28 @@
...
+    - `git commit -m "Release version X.Y.Z"`
+1. Tag the release:
+    - `git tag X.Y.Z`
+1. Push the changes in master branch and tags to the repository:
+    - `git push origin master`
+    - `git push origin X.Y.Z`
+1. Now the [PHP 5 version](https://github.com/lmc-eu/matej-client-php5) of the client needs to be updated and tagged as well:
+    1. Clone the repo: `git clone git@github.com:lmc-eu/matej-client-php5.git` or pull the master branch if you have it already cloned: `git pull origin master`
+    1. Inside the repository directory run transpilation script: `./transpile.sh`
+    1. Make sure the whole process completed successfully, including passed unit tests.
```
MortalFlesh, line 24: `Changelog` of the [PHP 5 version] should be updated as well.

MortalFlesh (changes requested):

RELEASE.MD
jirinovak: It's markup or github feature I guess. When you add some new bullets you have not manually renumber.

RELEASE.MD
jirinovak: It's not true, [PHP5 version] does not contain `Changelog` it instead link the one from this repo.

jirinovak (approved):

MortalFlesh (approved):

RELEASE.MD
OndraM: Every time I use this markdown numbering, someone asks this question :-D. It seems like not so know markdown feature 🤔.

Anyway you can check the markdown result here:  https://github.com/lmc-eu/matej-client-php/blob/30ee967130058c17d89bf0fd60cba3809d186285/RELEASE.MD

PR #92, opened 2018-11-05
Fix typo in readme

MortalFlesh:
There is some special char (obviously typo) in the `Exception` word


OndraM (approved):
Nice catch, thanks!

PR #96, opened 2019-01-17
Update exception handling to new Matej behavior

OndraM:
Invalid command inside recommendation/sorting request does not cause whole request to fail, rather the specific failed command status is `CommandResponse::STATUS_INVALID`.
The current Matej behavior (failing the whole Recommendation/Sorting request by its http response code) is inconsistent with Events endpoint and also with Matej own documentation ¯\\\_(ツ)_/¯.

The new behavior covered by this pull request will be part of upcoming Matej 7.56 (see RAD-2157 for the release and RAD-1806 for implementation details).

### DIFF
--- a/src/Model/CommandResponse.php
+++ b/src/Model/CommandResponse.php
@@ -10,6 +10,7 @@ use Lmc\Matej\Exception\ResponseDecodingException;
 class CommandResponse
 {
     public const STATUS_OK = 'OK';
+    /** @deprecated */
     public const STATUS_ERROR = 'ERROR';
     public const STATUS_SKIPPED = 'SKIPPED';
     public const STATUS_INVALID = 'INVALID';
--- a/src/Model/CommandResponse.php
+++ b/src/Model/CommandResponse.php
@@ -56,8 +56,15 @@ class CommandResponse
         return $this->data;
     }
 
+    /**
+     * Use this method to check whether this command response did not fail. Note both OK and SKIPPED statuses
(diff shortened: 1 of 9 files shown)


OndraM:
:warning: Just a small warning to everyone - if you relied on the old behavior (encapsulating sorting/recommendation request with try-catch to validate whether it was successful), you should update on new client version ASAP after Matej 7.56 is released and update your code accordingly (preferred way is to use $response->isSuccessful() method to validate the response - this method is being added in an [other pull request](https://github.com/lmc-eu/matej-client-php/pull/97))

src/Model/CommandResponse.php
```diff
@@ -57,6 +58,9 @@ public function getData(): array
 
     public function isSuccessful(): bool
     {
-        return $this->getStatus() === static::STATUS_OK;
+        // Both OK and SKIPPED command responses are marked as successful to allow easy "command did not fail" detection.
```
jirinovak: I would put these comments to method comment cause IDE show it when i try to call method.

jirinovak (dismissed):

src/Model/CommandResponse.php
OndraM: I'm not sure someone as a user of the library should be concerned about this... I'm seeing this more as a internal encapsulated implementation detail - this is why it wrote it inside, so that only someone interested in the implementation would came across it. 

But you think this should not be encapsulated and part of public phpdoc?

PR #97, opened 2019-01-17
Add isSuccessful() method to Response 

OndraM:
For simple check whether all contained command responses were successful.

This is also recommended way how to validate response successfulness.

If you don't need to process response (eg. when submitting events to Matej) you can now for example simply do this:

```php
public function submitDataToMatej(/*...*/): bool
    $response = $matej->request()
        ...
        ->send();
    
    // All commands were successful
    if ($response->isSuccessful()) {
        return true;
    }
    
    // At least one was not successful, so we log those
    foreach ($response->getCommandResponses() as $commandResponse) {
        if (!$commandResponse->isSuccessful()) {
            $this->logger->log('Command to Matej failed: ' . $commandResponse);
        }
    }
    
    return false;
}
```

### DIFF
--- a/tests/unit/Model/ResponseTest.php
+++ b/tests/unit/Model/ResponseTest.php
@@ -20,7 +20,8 @@ class ResponseTest extends UnitTestCase
         int $numberOfSuccessful,
         int $numberOfFailed,
         int $numberOfSkipped,
-        array $commandResponses
+        array $commandResponses,
+        bool $shouldBeSuccessful
     ): void {
         $responseId = uniqid((string) time());
         $response = new Response(
@@ -39,6 +40,8 @@ class ResponseTest extends UnitTestCase
 
         $this->assertSame($responseId, $response->getResponseId());
 
+        $this->assertSame($shouldBeSuccessful, $response->isSuccessful());
+
         $this->assertCount(count($commandResponses), $response->getCommandResponses());
         $this->assertContainsOnlyInstancesOf(CommandResponse::class, $response->getCommandResponses());
         for ($i = 0; $i < count($commandResponses); $i++) {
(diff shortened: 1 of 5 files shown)


tests/unit/Model/ResponseTest.php
```diff
@@ -56,10 +59,12 @@ public function provideResponseData(): array
         $skippedCommandResponse = (object) ['status' => CommandResponse::STATUS_SKIPPED, 'message' => '', 'data' => []];
 
         return [
-            'empty response data' => [0, 0, 0, 0, []],
-            'multiple successful commands' => [2, 2, 0, 0, [$okCommandResponse, $okCommandResponse]],
-            'multiple failed commands' => [2, 0, 2, 0, [$failedCommandResponse, $failedCommandResponse]],
-            'multiple skipped commands' => [2, 0, 0, 2, [$skippedCommandResponse, $skippedCommandResponse]],
+            'empty response data' => [0, 0, 0, 0, [], true],
```
jirinovak, line 62: Maybe `skipped and failed` case should be covered? 🤔

jirinovak (changes requested):

tests/unit/Model/ResponseTest.php
OndraM: In term of combinatorics - yes, but will it actually test anything - any path - that is not already tested?

There is just one combination missing to achieve all possible input combinations in this case, so here is it non-issue (and I added it). However in case of 4 inputs, would you test all 16 combinations (by manually written datasets in provider)? 🤔

🤔

danger-lmc:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Fails</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>There are some fixup/squash commits that needs to be applied before merge.</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e299fff93c1350a35c86b5880e6fa5c11dbd4fab
</p>

tests/unit/Model/ResponseTest.php
jirinovak: Yes you are right, my mistake.

Nope I wouldn't 😄

jirinovak (dismissed):

PR #98, opened 2019-02-07
Update cs, allow assert ^3.0

OndraM:


legendik (dismissed):

janicekt (dismissed):

OndraM:
Finally the BC check via phar is working (was in conflict with assertion library), and the build fail is expected (because of changes in Assertion library).

jirinovak (approved):

PR #113, opened 2020-03-29
Optional recommendation arguments

### LINKED ISSUE #108: Optional UserRecommendation arguments
Matej9 makes all scenario arguments except `userId` and `scenario` optional.

Remove arguments `count`, `rotationRate`, `rotationTime`, `responseProperties` from `UserRecommendation::create()`. It will accept exactly 2 arguments: `userId` and `scenario`. All other arguments can be provided using current fluent api:

```php
$response = $matej->request()
    ->recommendation(UserRecommendation::create('user-id', 'test-scenario')
        ->setCount(5)
        ->setRotationRate(0.3)
    )
    ->send();
```

This argument handling is already used for arguments `allowSeen` and `modelName`.
Some arguments setters already return instance of UserRecommendation that's being modified. Return it also from methods that currently don't: `setCount()`, `setScenario()`, `setRotationRate()`, `setRotationTime()`.

`UserRecommendation::UserRecommendation()` will not return arguments that were not set by the user and such arguments will not be send to Matej.

I considered using api similar to https://github.com/lmc-eu/matej-client-php/issues/107 but in this case it's not optimal. The fluent api is already used in this case and the arguments variable but known.

JakubTesarek:
BC BREAK

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ when providing recommendations.
 $matej = new Matej('accountId', 'apikey');
 
 $response = $matej->request()
-    ->recommendation(UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600))
+    ->recommendation(UserRecommendation::create('user-id', 'test-scenario'))
     ->setInteraction(Interaction::purchase('user-id', 'item-id')) // optional
     ->setUserMerge(UserMerge::mergeInto('user-id', 'source-id')) // optional
     ->send();
@@ -190,12 +190,14 @@ $response = $matej->request()
 $recommendations = $response->getRecommendation()->getData();
 ```
 
-You can also set more granular options of the recommendation command:
+You can also set more granular options of the recommendation command and overwrite Matej default behavior on per-request basis:
 
 ```php
-$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
-
-$recommendation->setFilters(['for_recommendation = 1'])
+$recommendation = UserRecommendation::create('user-id', 'test-scenario')
+    ->setCount(5)
+    ->setRotationRate(1.0)
+    ->setRotationTime(3600)
+    ->setFilters(['for_recommendation = 1'])
     ->setMinimalRelevance(MinimalRelevance::HIGH())
     ->enableHardRotation()
     ->addBoost(Boost::create('valid_to >= NOW()', 2));
@@ -235,13 +237,12 @@ Every item in Matej has its id, and optionally other item properties. These prop
 and you can upload item data in the [events](#send-events-data-to-matej) request. This has major benefit because you can request
 these properties to be returned as part of your Recommendation Request.
 
-We call them response properties, and they can be specified either as the last parameter of `UserRecommendation::create` function,
-by calling `->addResponseProperty()` method, or by calling `->setResponseProperties()` method. Following will request an `item_id`,
-`item_url` and `item_title`:
+We call them response properties. They can be specified by calling `->addResponseProperty()` method or by calling `->setResponseProperties()` method. Following will request an `item_id`, `item_url`, `item_title`:
 
 ```php
-$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600, ['item_url'])
-    ->addResponseProperty('item_title');
+$recommendation = UserRecommendation::create('user-id', 'test-scenario')
+    ->addResponseProperty('item_title')
+    ->addResponseProperty('item_url');
 
 $response = $matej->request()
     ->recommendation($recommendation)
@@ -320,7 +321,7 @@ $response = $matej->request()
     ->addSorting(Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3']))
     ->addSortings([/* array of Sorting objects */])
     // Request user-based recommendations
-    ->addRecommendation(UserRecommendation::create('user-id', 10, 'emailing', 1.0, 3600))
+    ->addRecommendation(UserRecommendation::create('user-id', 'emailing'))
     ->addRecommendations([/* array of UserRecommendation objects */])
     ->send();
 ```
@@ -332,11 +333,11 @@ but once available, you can specify which model you want to use when requesting 
 This is available for `recommendation`, `sorting` and `campaign` requests:
 
 ```php
-$recommendationCommand = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
-$recommendationCommand->setModelName('alpha');
+$recommendationCommand = UserRecommendation::create('user-id', 'test-scenario')
+    ->setModelName('alpha');
 
 $sortingCommand = Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3']);
-$sortingCommand->setModelName('beta')
+$sortingCommand->setModelName('beta');
 
 $response = $matej->request()
     ->recommendation($recommendationCommand)
@@ -359,7 +360,7 @@ Typically, you'd select a random sample of users, to which you'd present recomme
 in your code should look similar to this:
 
 ```php
-$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
+$recommendation = UserRecommendation::create('user-id', 'test-scenario');
 
 if ($session->isUserInBucketB()) {
     $recommendation->setModelName('alpha');
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -1,0 +1,34 @@
+# Upgrading from 2.x to 3.0
+
+API client release 2.0 contains few backward incompatible changes.
+
+This guide will help you upgrade your codebase.
+
+## `UserRecommendation::create()` now accepts only `$user_id` and `$scenario` 
+`UserReccomentation::create()` accepts only two argumens: `$user_id` and `$scenario`.
+Both arguments are required. All other arguments are optional their default values are
+configured using admin 
+
+Recommendation command can be further parametrized using fluent API.
+
+#### Before
+```php
+$recommendation = UserRecommendation::create('user-id', 5, 'scenario', 1.0, 3600);
+```
+
+#### After
+```php
+$recommendation = UserRecommendation::create('user-id', 'scenario')
+	->setCount(5)
+	->setRotationRate(1.0)
+	->setRotationTime(3600);
+```
+
+which is equivalent to
+
+```php
+$recommendation = UserRecommendation::create('user-id', 'scenario');
+$recommendation->setCount(5);
+$recommendation->setRotationRate(1.0);
+$recommendation->setRotationTime(3600);
+```
--- a/src/Model/Command/UserRecommendation.php
+++ b/src/Model/Command/UserRecommendation.php
@@ -25,62 +25,37 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
     /** @var int */
     private $rotationTime;
     /** @var bool */
-    private $hardRotation = false;
+    private $hardRotation;
     /** @var MinimalRelevance */
     private $minimalRelevance;
     /** @var string[] */
-    private $filters = [];
+    private $filters;
     /** @var string */
     private $filterType = self::FILTER_TYPE_MQL;
     /** @var string|null */
     private $modelName;
     /** @var string[] */
-    private $responseProperties = [];
+    private $responseProperties;
     /** @var bool */
     private $allowSeen = false;
     /** @var Boost[] */
     private $boosts = [];
 
-    private function __construct(
-        string $userId,
-        int $count,
-        string $scenario,
-        float $rotationRate,
-        int $rotationTime,
-        array $responseProperties
-    ) {
-        $this->minimalRelevance = MinimalRelevance::LOW();
-
+    private function __construct(string $userId, string $scenario)
+    {
         $this->setUserId($userId);
-        $this->setCount($count);
         $this->setScenario($scenario);
-        $this->setRotationRate($rotationRate);
-        $this->setRotationTime($rotationTime);
-        $this->setResponseProperties($responseProperties);
     }
 
     /**
      * @param string $userId
-     * @param int $count Number of requested recommendations. The real number of recommended items could be lower or
-     * even zero when there are no items relevant for the user.
      * @param string $scenario Name of the place where recommendations are applied - eg. 'search-results-page',
      * 'emailing', 'empty-search-results, 'homepage', ...
-     * @param float $rotationRate How much should the item be penalized for being recommended again in the near future.
-     * Set from 0.0 for no rotation (same items will be recommended) up to 1.0 (same items should not be recommended).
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -339,8 +345,11 @@ but once available, you can specify which model you want to use when requesting
 This is available for `recommendation`, `sorting` and `campaign` requests:
 
 ```php
-$recommendationCommand = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
-$recommendationCommand->setModelName('alpha');
+$recommendationCommand = UserRecommendation::create('user-id', 'test-scenario');
```
jirinovak:
```suggestion
$recommendationCommand = UserRecommendation::create('user-id', 'test-scenario')
```

README.md
```diff
@@ -366,7 +375,7 @@ Typically, you'd select a random sample of users, to which you'd present recomme
 in your code should look similar to this:
 
 ```php
-$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
+$recommendation = UserRecommendation::create('user-id', 'test-scenario')
```
jirinovak:
```suggestion
$recommendation = UserRecommendation::create('user-id', 'test-scenario');
```

jirinovak (commented):
@VincTheSecond told me about this future change last week. I checked the changes to have better idea what it means for us in jobs.cz. I leave few typo comments and hope you don't mind 😄

README.md
```diff
@@ -227,18 +230,26 @@ $recommendations = $response->getRecommendation()->getData();
...
 these properties to be returned as part of your Recommendation Request.
 
-We call them response properties, and they can be specified either as the last parameter of `UserRecommendation::create` function,
-by calling `->addResponseProperty()` method, or by calling `->setResponseProperties()` method. Following will request an `item_id`,
+We call them response properties. They can be specified by calling `->addResponseProperty()` method or by calling `->setResponseProperties()` method. Following will request an `item_id`,
 `item_url` and `item_title`:
 
 ```php
-$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600, ['item_url'])
+$recommendation = UserRecommendation::create('user-id', 'test-scenario')
```
OndraM, line 243: It is not clear why this will return "`item_id`, `item_url` and `item_title`" as stated in the paragraph above. How do you achieve such behavior (ie. returning item_url)?

README.md
```diff
@@ -318,8 +329,12 @@ $response = $matej->request()
     ->addSorting(Sorting::create('user-id', ['item-id-1', 'item-id-2', 'item-id-3']))
     ->addSortings([/* array of Sorting objects */])
     // Request user-based recommendations
-    ->addRecommendation(UserRecommendation::create('user-id', 10, 'emailing', 1.0, 3600))
-    ->addRecommendations([/* array of UserRecommendation objects */])
+    ->addRecommendation(
+        UserRecommendation::create('user-id', 'emailing')
+            ->setCount(10)
```
OndraM: I think the setters could be omitted here to make the example shorter. Is should just show hot to pass `UserRecommendation`, no need to set the optional properties.

README.md
```diff
@@ -330,8 +345,11 @@ but once available, you can specify which model you want to use when requesting
 This is available for `recommendation`, `sorting` and `campaign` requests:
 
 ```php
-$recommendationCommand = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
-$recommendationCommand->setModelName('alpha');
+$recommendationCommand = UserRecommendation::create('user-id', 'test-scenario');
+    ->setCount(5)
+    ->setRotationRate(1.0)
+    ->setRotationTime(3600)
```
OndraM: Dtto, would be IMHO better to omit those optional setters on leave there just `setModelName`, which is what the example is all about.

UPGRADE-3.0.md
```diff
@@ -0,0 +1,32 @@
+# Upgrading from 2.x to 3.0
+
+API client release 2.0 contains few backward incompatible changes.
+
+This guide will help you upgrade your codebase.
+
+## `UserRecommendation::create()` now accepts only `$user_id` and `$scenario` 
+`UserReccomentation::create()` accepts only two argumens: `$user_id` and `$scenario`.
+Both are arguments are required. Reccomendation command can be further parametrized
```
OndraM:
```suggestion
Both are arguments are required. Recommendation command can be further parametrized
```

README.md
```diff
@@ -193,8 +193,11 @@ $recommendations = $response->getRecommendation()->getData();
 You can also set more granular options of the recommendation command:
 
 ```php
-$recommendation = UserRecommendation::create('user-id', 5, 'test-scenario', 1.0, 3600);
-$recommendation->setFilters(['for_recommendation = 1'])
+$recommendation = UserRecommendation::create('user-id', 'test-scenario')
+    ->setCount(5)
```
OndraM, line 197: Is it possible to explain what is the default behavior and what happens if the setters are used? 

I mean something like: 

> Default behavior is specified in Matej admin console, but it could be overridden on per-request basis using `setCount()`, `setRotationRate()` etc. methods of `UserRecommendation`.

UPGRADE-3.0.md
```diff
@@ -0,0 +1,32 @@
...
+using fluent API.
```
OndraM: Also here could be mentioned that default parameters are now being set using Matej Admin Console and setters only override them.

OndraM:
Rebased on current master (with merged boosting rules)

src/Model/Command/UserRecommendation.php
```diff
@@ -41,46 +41,22 @@ class UserRecommendation extends AbstractCommand implements UserAwareInterface
...
-        string $userId,
-        int $count,
-        string $scenario,
-        float $rotationRate,
-        int $rotationTime,
-        array $responseProperties
-    ) {
+    private function __construct(string $userId, string $scenario)
+    {
         $this->minimalRelevance = MinimalRelevance::LOW();
```
OndraM: This value of minimalRelevance will be always sent as part of the request?

src/Model/Command/UserRecommendation.php
```diff
@@ -137,7 +113,11 @@ public function setFilters(array $filters): self
     }
 
     /**
-     * Add another response property you want returned. item_id is always returned by Matej.
+     * Add another response property you want returned. item_id is always
+     * returned by Matej.
```
OndraM: No need to break line here

src/Model/Command/UserRecommendation.php
```diff
@@ -227,11 +207,20 @@ protected function setUserId(string $userId): void
         $this->userId = $userId;
     }
 
-    protected function setCount(int $count): void
+    /**
+     * Set number of requested recommendations. The real number of recommended
```
OndraM: Line limit is 120 characters, no need to break line on 80.

src/Model/Command/UserRecommendation.php
```diff
@@ -241,18 +230,36 @@ protected function setScenario(string $scenario): void
         $this->scenario = $scenario;
     }
 
-    protected function setRotationRate(float $rotationRate): void
+    /**
+     * Set how much should the item be penalized for being recommended again in
+     * the near future.
```
OndraM: Dtto, could be one line

ghost:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Fails</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Backward incompatible changes detected</td>
    </tr>
  </tbody>
</table>




## Backward incompatible changes introduced by this PR
 - [BC] The parameter $count of Lmc\Matej\Model\Command\UserRecommendation::create() changed from int to a non-contravariant string
 - [BC] The parameter $count of Lmc\Matej\Model\Command\UserRecommendation::create() changed from int to string
 - [BC] The return type of Lmc\Matej\Model\Command\UserRecommendation#setCount() changed from void to the non-covariant self
 - [BC] The return type of Lmc\Matej\Model\Command\UserRecommendation#setCount() changed from void to self
 - [BC] The return type of Lmc\Matej\Model\Command\UserRecommendation#setRotationRate() changed from void to the non-covariant self
 - [BC] The return type of Lmc\Matej\Model\Command\UserRecommendation#setRotationRate() changed from void to self
 - [BC] The return type of Lmc\Matej\Model\Command\UserRecommendation#setRotationTime() changed from void to the non-covariant self
 - [BC] The return type of Lmc\Matej\Model\Command\UserRecommendation#setRotationTime() changed from void to self


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 691ff99d57def30109aa4a4e8d1a18c286717136
</p>

src/Model/Command/UserRecommendation.php
```diff
@@ -279,17 +286,26 @@ protected function getCommandParameters(): array
...
-            'rotation_rate' => $this->rotationRate,
-            'rotation_time' => $this->rotationTime,
             'hard_rotation' => $this->hardRotation,
             'min_relevance' => $this->minimalRelevance->jsonSerialize(),
             'filter' => $this->assembleFiltersString(),
             'filter_type' => $this->filterType,
             'properties' => $this->responseProperties,
         ];
 
+        if ($this->count !== null) {
```
OndraM, line 295: What about other params? hard_rotation, min_relevance, filter, filter_type, properties

They will always be added to the request, even if not set by client. Is this intentional?

tests/unit/Model/Command/UserRecommendationTest.php
```diff
@@ -10,18 +10,15 @@ class UserRecommendationTest extends TestCase
...
             [
                 'type' => 'user-based-recommendations',
                 'parameters' => [
                     'user_id' => 'user-id',
-                    'count' => 333,
                     'scenario' => 'test-scenario',
-                    'rotation_rate' => 1.0,
-                    'rotation_time' => 3600,
                     'hard_rotation' => false,
                     'min_relevance' => MinimalRelevance::LOW,
```
OndraM: I may be wrong, but I expected only user_id and scenario in the request, as these are the only parameters set for the instance of UserRecommendation. Why are there some defaults?

OndraM (changes requested):

README.md
JakubTesarek: fixed

README.md
JakubTesarek: fixed

README.md
JakubTesarek: It was a mistake in the example code. Btw. we will always return item_id and technically it's not a item property. We are still not sure what should matej API do if user requests it. But that's out of scope of this PR.

README.md
JakubTesarek: ok

README.md
JakubTesarek: ok

UPGRADE-3.0.md
JakubTesarek: fixed

README.md
JakubTesarek: Ok but I would like to point out that so far the readme didn't mention anything that could be set using admin console and it was expected that the user understands how to use the API. I don't really see how this is different.

JakubTesarek: That's not technically true. They can't be set using admin console. If they are not set, Matej will use some defaults but they are not configurable using the admin console. I assumed the reader implicitly understands that if they don't configure a value manually, Matej uses some default. So in this case I would have to write something like "You can set some values manually and if you don't Matej uses some defaults.". That's quite obvious and also it's not documenting behaviour of this library but behaviour of Matej which should be described in Mate documentation.

UPGRADE-3.0.md
JakubTesarek: Same as above

README.md
JakubTesarek: Btw. this is exactly the same behaviour as before. For example minimalRelevance was always optional argument configured by setters and from the point of view of user of this library used default value if not set.

src/Model/Command/UserRecommendation.php
JakubTesarek: Thanks for pointing that out. I totally missed all the other attributes that are optional

src/Model/Command/UserRecommendation.php
JakubTesarek: fixed

src/Model/Command/UserRecommendation.php
JakubTesarek: fixed

src/Model/Command/UserRecommendation.php
JakubTesarek: fixed

src/Model/Command/UserRecommendation.php
JakubTesarek: No, it's an oversight. Fixed

tests/unit/Model/Command/UserRecommendationTest.php
JakubTesarek: You are correct. They're here because I goofed

JakubTesarek:
Fixes implemented. @OndraM

README.md
OndraM: But as someone using the library, I would consider important to know what happens, if I don't specify the values on per-request basis. Whether some default defined in the library will be used, whether this is internal on Matej side etc.

What changing the line before this example to something like this:

```
You can also set more granular options of the recommendation command and overwrite Matej default behavior on per-request basis:
```

(I thought from our discussion those values would be configurable by users in the admin consele - I probably got it wrong :-) )

src/Model/Command/UserRecommendation.php
```diff
@@ -180,7 +165,7 @@ public function setModelName(string $modelName): self
     /**
      * Allow items, that the user has already "seen"
      *
-     * By default user won't see any items, that it has visitted (and we have recorded DetailView interaction.)
+     * By default user won't see any items, that it has visitted (and we have recorded DetailView  interaction.)
```
OndraM: Typo (was present before)

```suggestion
     * By default user won't see any items, that it has visited (and we have recorded DetailView  interaction.)
```

src/Model/Command/UserRecommendation.php
```diff
@@ -227,11 +212,19 @@ protected function setUserId(string $userId): void
...
+    public function setCount(int $count): self
     {
         Assertion::greaterThan($count, 0);
 
         $this->count = $count;
+
+        return $this;
     }
 
     protected function setScenario(string $scenario): void
```
OndraM: We tend to order methods by visibility, ie. public > private > protected - so `setUserId()` and `setScenario()` should be after all public methods.

OndraM (changes requested):

src/Model/Command/UserRecommendation.php
JakubTesarek: fixed

src/Model/Command/UserRecommendation.php
JakubTesarek: fixed

OndraM (approved):

PR #118, opened 2020-08-18
Removed unused filter_type parameter.

JakubTesarek:
All tests passed against matej-client-test


VincTheSecond (approved):

OndraM:
Even though unused, removing of public property is BC break and must not be done as part of non-major releases.

The BC check build was broken (#122), so this was not reported by the CI.

OndraM:
Fixed in #122.

PR #123, opened 2020-09-21
Fixed bug when recommendations were returned as strings.

JakubTesarek:
This fixes a bug caused by special behaviour of Matej.

When sending RecommendationCommand, Matej returns one of 2 types of responses. If the request or used scenario specifies item properties to get, it returns list of objects where there's always a field `item_id` present. This is true even if the list of properties is empty.

For backwards compatibility, Matej returns list of strings instead of objects if the request doesn't contain `item_properties` field and no scenario is used.

When Matej collects fields to return, it first looks into the request and if there's no `item_properties` field, it looks into the configured scenario. In version 2.* matej-client-php used to always send this field so users never experienced this bug. in #113 we removed that argument and we send it only when user specifies item properties to get.

This is a temporary bugfix until Matej unify the behaviour of the recommendation request. It will not be necessary afterwards.


VincTheSecond (dismissed):
Otestoval jsem vsechny mozne kombinace properties vs. konfiguraci scenaru a klient vzdy spravne vratil pole dictu. ✅

Na samotny commit review udelat neumim. :-)

src/Model/RecommendationCommandResponse.php
```diff
@@ -0,0 +1,19 @@
+<?php declare(strict_types=1);
+
+namespace Lmc\Matej\Model;
+
+class RecommendationCommandResponse extends CommandResponse
+{
+    public function getData(): array
+    {
+        $data = [];
+        foreach ($this->data as $key => $val) {
```
OndraM, line 17: A comment explaining why this conversion is being done might be handy

src/Model/Response/RecommendationsResponse.php
```diff
@@ -3,22 +3,41 @@
...
     {
-        return $this->getCommandResponse(2);
+        return $this->getCommandResponse(static::RECOMMENDATION_INDEX);
+    }
+
+    protected function decodeRawCommandResponses(array $commandResponses): array
+    {
+        $decodedResponses = [];
+        foreach ($commandResponses as $index => $rawCommandResponse) {
+            if ($index == static::RECOMMENDATION_INDEX) {
```
OndraM:
```suggestion
            if ($index === static::RECOMMENDATION_INDEX) {
```

tests/unit/Model/Response/RecommendationsResponseTest.php
```diff
@@ -43,6 +43,56 @@ public function shouldBeInstantiable(): void
...
+        ];
+        $userMergeCommandResponse = (object) [
+            'status' => CommandResponse::STATUS_OK,
+            'message' => 'MOCK_USER_MERGE_MESSAGE',
+            'data' => ['MOCK' => 'USER_MERGE'],
+        ];
+        $recommendationCommandResponse = (object) [
+            'status' => CommandResponse::STATUS_OK,
+            'message' => 'MOCK_RECOMMENDATION_MESSAGE',
+            'data' => ['MOCK_ITEM_ID'],
```
OndraM: There is a lot of copy-paste in those two tests - in fact, this is the only line in which they are different, isn't it? I'd suggest either extracting the common logic to some helper methods or maybe better create data provider which will provide value of `data` here (ie. once it will provide object, once it will provide plain array as here).

OndraM (changes requested):
Also please don't forget to look at those codestyle violations reported by the CI.

JakubTesarek:
> Also please don't forget to look at those codestyle violations reported by the CI.

What violations? I didn't get any violations locally and CI report is all green except one that ran out of memory.

OndraM:
> What violations? I didn't get any violations locally and CI report is all green except one that ran out of memory.

Yup, sorry, you are right - there are no errors in this run, however the codestyle check was not working properly (I already fixed in master), so it will report some issues once you rebase this branch with master and run `composer update`. For example there are few missing newlines at the end of files etc.

OndraM (approved):
👍

PR #126, opened 2020-10-09
Change dash in item-id property name to underscore for backward compatibility

hokypierce:
I found a bug in https://github.com/lmc-eu/matej-client-php/pull/123 which caused that we wasnt able to process response from matej client.


OndraM (approved):

JakubTesarek:
Missing entry in changelog. I'll add it when merging.

JakubTesarek (approved):

PR #131, opened 2020-11-29
Added support for new recommendation commands.

JakubTesarek:
- Added support for new recommendation commands.
- Removed PhpStan. Mainly because it doesn't work. It doesn't understand that classes have access to methods from other classes and doesn't support Traits.
- Removed deprecated constant FILTER_TYPE_MQL

Integration tests are failing because this change expects new features in Matej which are not yet in production.


OndraM:
PHPStan is essential tool to prevent many unwanted bugs. We use it successfully in all other PHP projects, what is the issue here? I'm *strongly* in favor of fixing the issues instead of removing PHPStan check.

OndraM:
Also, wouldn't be possible to split the pull request to few smaller ones? :pray: 

Its quite hard to review 1000+ lines of code in 30+ files... Usual rule of thumb is to keep PR with less than ~250 lines changed (with some exceptions in some edge cases), so that reviewers are cognitively able to review it properly.

JakubTesarek:
@OndraM 
There are mainly 2 issues. 1. Every time I run PhpStan it reports something else. 2. it doesn't resolve inheritance. I have two classes:

```
abstract class AbstractKlass {
    protected function __construct() {}
}

class ConcreteKlass extends AbstractKlass {
    public function someFunction() {}
    public static function create(): self
    {
        return new static();
    }
}
```

If I construct new instance like this `$ck = ConcreteKlass::create()` and attempt to call it `$ck->someFunction()` it works as expected. But PhpStan reports error that I'm calling unknown function `someFunction` on abstract class `AbstractKlass`. In this case `$ck` is obviously instance of `ConcreteKlass` and even the tests show it.

JakubTesarek:
@OndraM I never heard of this `rule of thumb` concerning number of lines in PR. In my opinion every PR should contain only the things but all the things that are necessary for the feature to work. There is no way to split this into smaller PR that would still work. All of this is one feature. We have to have it all or not at all.

I can split out a PR for removing the deprecated constant and maybe on for the PhpStan but that's 2 lines of code and I don't see how would that help.

OndraM:
> Every time I run PhpStan it reports something else. 

Do you have an example? There is caching involved, so there *hypothetically* may be a bug, maybe try upgrading to latest phpstan version?

> But PhpStan reports error that I'm calling unknown function someFunction on abstract class AbstractKlass. 

I tried this in PHPStan playground, it seems to be working as expected, there is no such error: https://phpstan.org/r/d0446f4f-13f3-47a1-9717-719174258223 Do you have some other example? 

I strongly advocate to keep PHPStan, its one of the most important tools to keep the code sane. Removing it is like removing unit tests because they are not passing... Yes, it may find some false-positives, but adding them to ignore list is much cheaper than dealing with issues in production.

tests/integration/RequestBuilder/EventsRequestBuilderTest.php
```diff
@@ -71,12 +75,39 @@ public function shouldExecuteInteractionAndUserMergeAndItemPropertyCommands(): v
...
+
+            $properties = [];
+            foreach ($resp->getData() as $property) {
+                $properties[] = $property->name;
+            }
+
+            if (!array_diff(static::getPropertiesList(), $properties)) {
+                return;
+            }
+            usleep(1000000); # 0.1s
```
kdosiodjinud, line 101: 1M in usleep is 1 second :)

kdosiodjinud (approved):

PR #135, opened 2021-02-26
Update new coding standard & other dependencies

OndraM:
Update done according to https://github.com/lmc-eu/php-coding-standard/blob/main/UPGRADE-3.0.md .

However this needs PHP 7.3 - I'm not sure if we can bump the dependency or not yet?


kdosiodjinud (approved):

.github/workflows/php.yaml
```diff
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: ['7.1', '7.2', '7.3', '7.4']
+        php-version: ['7.3', '7.4']
```
jakubpilaralmamedia, line 16: tady by se mohla přidat i nějaká 8.x verze, ne?

jakubpilaralmamedia (approved):

.github/workflows/php.yaml
OndraM: See #143. Done in separate PR, to keep this one focused on one thing (ie. to get it working with the new coding standard).

PR #138, opened 2021-11-18
Items filtering trait: Convert boosts array to list.

janbarasek:
Just fix bug from Seduo.

Matej should convert the boost array to a list if it does not contain properly indexed keys.

### DIFF
--- a/src/Model/Command/ItemsFilteringTrait.php
+++ b/src/Model/Command/ItemsFilteringTrait.php
@@ -82,10 +82,11 @@ trait ItemsFilteringTrait
 
     /**
      * Set boosts. Removes all previously set rules.
+     * @param Boost[] $boosts
      */
     public function setBoosts(array $boosts): self
     {
-        $this->boosts = $boosts;
+        $this->boosts = array_values($boosts);
 
         return $this;
     }


src/Model/Command/ItemsFilteringTrait.php
```diff
@@ -82,10 +82,11 @@ public function addBoost(Boost $boost): self
 
     /**
      * Set boosts. Removes all previously set rules.
+     * @param Boost[] $boosts
      */
     public function setBoosts(array $boosts): self
     {
```
OndraM, line 88: Best would also be to add something like
```php
Assertion::allIsInstanceOf($boosts, Boost::class);
```
right on beginning of the method body.

And maybe then is the `array_values(...)` unnecessary, isn't it? Or in what case did this issue occur?

OndraM (changes requested):
Thanks @janbarasek, nice catch!

Would be possible to add a test (or extend a existing one) for this case?

(However if the change would be just in the assertion, as in my suggestion, then the tests for just an type assertion would not be needed.)

petrofcikmatus:
Hi @janbarasek, It has been 22 days since you created this PR. Do you plan to complete this issues? Thanks, M.

janbarasek:
This PR is no longer relevant for me and I solved it in a different way.

src/Model/Command/ItemsFilteringTrait.php
janbarasek: The `array_values()` call was very important when the previous application layer implemented a filter to remove certain array items by key.

The `array_values()` function then created a list again from the filtered array.

PR #141, opened 2024-01-04
Replace HTTP Factory class

jakubpilaralmamedia:
Fix: Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead.


composer.json
```diff
@@ -21,14 +21,15 @@
...
         "php-http/promise": "^1.0",
+        "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0",
         "ramsey/uuid": "^3.7 || ^4.0"
     },
     "require-dev": {
         "ergebnis/composer-normalize": "^2.4",
-        "lmc/coding-standard": "^1.3 || ^2.0",
+        "http-interop/http-factory-guzzle": "^1.2",
+        "lmc/coding-standard": "^1.3 || ^2.0 || ^3.0",
```
OndraM: Its not possible to make it compatible to all these versions, because there are incompatible changes in 3.0 (especially config format changed to PHP).

I think https://github.com/lmc-eu/matej-client-php/pull/135 should be resurrected and finished first. Are you able to take over the PR #135 and finish it?

OndraM (changes requested):

composer.json
```diff
@@ -21,14 +21,15 @@
...
-        "php-http/message-factory": "^1.0",
         "php-http/promise": "^1.0",
+        "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0",
         "ramsey/uuid": "^3.7 || ^4.0"
     },
     "require-dev": {
         "ergebnis/composer-normalize": "^2.4",
-        "lmc/coding-standard": "^1.3 || ^2.0",
+        "http-interop/http-factory-guzzle": "^1.2",
```
petrofcikmatus: Shouldn't be required oldest possible version with needed features?
```suggestion
        "http-interop/http-factory-guzzle": "^1.0",
```
I'm affraid that `^1.2` could be problem if some projects use `1.1.x` or `1.0.x` versions.

src/Matej.php
```diff
@@ -27,7 +28,7 @@ public function request(): RequestBuilderFactory
     }
 
     /** @return $this */
-    public function setHttpClient(HttpClient $client): self
+    public function setHttpClient(ClientInterface $client): self
```
petrofcikmatus, line 31: All these new types look like breaking changes. Are you gonna release new major version? If so, there is no minor version with deprecation notices for easier transition?

petrofcikmatus (changes requested):

composer.json
jakubpilaralmamedia: Yes, it's ok to change it in #135, I don't need it now. It's because of PHP8.3 need it.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 <!-- There is always Unreleased section on the top. Subsections (Added, Changed, Fixed, Removed) should be added as needed. -->
 
 ## Unreleased
+### Added
+- Fix: Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead.
```
OndraM: This is not a fix. It replaces underlying HTTP library interfaces, meaning the signatures of public methods will change. This is unfortunately a breaking change.

This also means there should be an UPGRADE document prepared showing a steps to upgrade (see root directory of the repository for previous upgrade document).

src/Matej.php
OndraM: I don't think it would be possible / effective to have this transition path with deprecation notices. IMHO releasing new version with new PHP verison dependencies would be enough...

jakubpilaralmamedia: I think, It's not breaking change, because old HttpClient can be still used, becuase it implements this interface - https://github.com/php-http/httplug/blob/2.x/src/HttpClient.php

CHANGELOG.md
jakubpilaralmamedia: But there is no changes needed. It will works without changes 🤷

tests/unit/Http/RequestManagerTest.php
```diff
@@ -26,8 +26,29 @@ public function shouldSendAndDecodeRequest(): void
             __DIR__ . '/Fixtures/response-one-successful-command.json'
         );
 
-        $mockClient = new Client();
-        $mockClient->addResponse($dummyHttpResponse);
+        $mockClient = $this->createMock(Psr18Client::class);
```
OndraM, line 29: I think the Mock client, which is implementation specifically made for unit tests, should be used instead. It is also already part of require-dev.

The assertions are then more readable and understandable then defining mocks like this.

See https://docs.php-http.org/en/latest/clients/mock-client.html

tests/unit/MatejTest.php
```diff
@@ -25,8 +26,10 @@ public function shouldExecuteRequestViaBuilder(): void
             __DIR__ . '/Http/Fixtures/response-one-successful-command.json'
         );
 
-        $mockClient = new Client();
-        $mockClient->addResponse($dummyHttpResponse);
+        $mockClient = $this->createMock(Psr18Client::class);
```
OndraM, line 29: Dtto, use directly the mock client.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 ## Unreleased
 ### Changed
 - Require PHP ^7.3
+- Drop old version php-http/client-common ^1.6
```
OndraM, line 10:
```suggestion
- Allow only `php-http/client-common` ^2.0 dependency (drop ^1.6)
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
...
+
+### Added
+- Replace abandoned package php-http/message-factory. Using psr/http-factory instead.
```
OndraM, line 13:
```suggestion
- Replace dependency on abandoned `php-http/message-factory` with `psr/http-factory`.
```

OndraM (changes requested):

